PR #122, opened 2017-03-20
Losses calculators

rmroot:
Updates to number field inputs and charge material saving


rmroot:
Wall losses commits added. 

Wall losses are calculating and saving the same as charge material.

Wall Loss example from unit tests found under Mock PHAST 1

PR #138, opened 2017-03-28
Phast design refactor

rmroot:
Round and Quad Opening losses working with "unit test" added to Mock Phast 1

Will need to `npm install` to get latest `amo-tools-suite`


rmroot:
Cooling losses added

rmroot:
Fixture losses added with "unit tests" working in Phast Mock 1,

PR #176, opened 2017-04-12
Update modal

kbeanblossom:
Added progress indicator while update is downloading. Application will automatically quit and bring up the installer when the update download is complete.


rmroot:
@kbeanblossom do we need autoUpdater imported to core.module ?

rmroot:
Merging. Will clean up later if unneeded import

PR #328, opened 2017-06-06
Phast updates

rmroot:
Global PHAST settings (electric or fuel fired)
Slag Losses
Auxilary Losses
PHAST Styling consistent with PSAT


rmroot:
Auto save branch was merged into phast updates

PR #394, opened 2017-06-19
Added underline to Input and Description on all the help panels. Adde…

vanced:
…d text according to Kristina's Suggestions. Should only be text and formatting changes.


roleri17:
Bien

PR #416, opened 2017-06-26
Super scripts

RoundDev:
Super scripts for unit converter.


vanced:
Have lost the acronyms for units without superscripts on the convert units calculator. Missing acronym for cubic feet per meter.

PR #442, opened 2017-07-07
fix for explore opportunities unit bug

rmroot:


rmroot:
fixed bug #438 and bug #441

Issue #484, opened 2017-07-17
PSAT report updates


rmroot:

rmroot:
I am going to need the input field values that correspond with an optimized setup from the tools-suite to be able to implement the "Optimize" button

PR #498, opened 2017-07-24
Tools suite update

rmroot:
Updated desktop to use latest tools-suite

Report updates

Explore opportunities updates

Using optimized or existing results


rmroot:
Will need to do `npm install` after merge to update tools-suite

rmroot:
issue #484

PR #606, opened 2017-08-29
small detail fixes to psat

rmroot:


rmroot:
Fixes default values for psat

Fixes -00 for reports

Fixes error msg with rated and field motor power

Issue #643, opened 2017-09-07
PHAST - EAF - Energy Input


koay9f:
Some changes to the EAF - Energy Input page

Working with Preston on the C++ side.  Here's updated notes for the UI side.

![image](https://user-images.githubusercontent.com/25389736/30183370-33862a0c-93e8-11e7-968d-b1860ebe9da1.png)

koay9f:
This has a branch with the C++ side finished.

Issue #639, opened 2017-09-07
PHAST - EAF - Exhaust Gases


koay9f:
Small error fix & change what is being asked for & what is being returned.

src/calculator/losses/ExhaustGasEAF.cpp
Typo:Line 22 -  1000 should be 1020
![image](https://user-images.githubusercontent.com/25389736/30173318-383d6e16-93c5-11e7-94a4-ab3bfcbdf89c.png)

Line 60 - remove 
this->cycleTime_ and this->otherLosses_ from this equation
we want the result to be per hour not per cycle and we are going to ask for other losses later.

bindings/phast.h
line 142 - change to
`double heatLoss = eg.getTotalHeatFlue();`
we want it to be returned in btu/hr not kW

Now cycleTime_ and otherLosses_ are no longer being used on **this page** and can be removed from where all it needs to be here.

koay9f:
There is a branch with the C++ side finished

PR #645, opened 2017-09-11
Tools suite update

rmroot:
Electric Arc Furnace to match new tools-suite

Models and services related to ExhaustGas and EnergyInput updated to reflect changes:

exhaustGas method is now exhaustGasEAF
energyInput method is now energyInputEAF

measuredOxygenFlow and naturalGasFlow are gone from the expected inputs for energyInputEAF

cycleTime, O2 and CO2 are gone from the inputs for exhaustGasEAF

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@angular/platform-server": "^4.2.1",
     "@angular/router": "^4.2.1",
     "ajv": "^5.2.2",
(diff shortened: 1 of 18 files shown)


rmroot:
issue #643

rmroot:
issue #639

Issue #681, opened 2017-09-21
PHAST - Displayed Results


koay9f:
For all PHAST results (even within the calculators)- show no decimal places

FYI - check the Sankey, it might already be MMBtu/hr

PR #710, opened 2017-09-26
Phast updates

rmroot:
reworked validation check for phast calculations to set results to null if invalid and show '- -' in place of zero/null values


rmroot:
issues #709 and #681

PR #718, opened 2017-09-27
pump curve tab and calculator added to psat.

rmroot:


rmroot:
calculator added, no data binding to it yet. Still need information on how we want that to work

rmroot:
issue #716

Issue #735, opened 2017-10-05
PHAST - steam - HHV


koay9f:
Use "total heat for steam" from design or metered energy use for summary HHV

if metered is filled out (> 0 ?) then use that

if multiple zones for design energy use, average the "total heat for steam" across all zones.

PR #737, opened 2017-10-06
Phast updates

rmroot:
Updates to PHAST Report:
Graphs can no choose which modification you would like to compare
Sankey can also choose which modification you would like to compare


rmroot:
closes issue #735

PR #741, opened 2017-10-10
Flue Gases Help Field by Field 

RoundDev:
Flue Gases Help Field by Field for Add New Gas Modal


rmroot:
Looks good, I updated the CSS for the help

Issue #746, opened 2017-10-11
PHAST - Update Modify Conditions Tabs


rmroot:
Indicate required tabs,
Add operations tab and content

PR #749, opened 2017-10-12
operations tab hooked up

rmroot:
Part two of issue #746 still need to re-order tabs and update styles


rmroot:
Also adds Payback period and implementation costs to assessment inputs and Executive Summary

Issue #783, opened 2017-10-18
Unit Converter - Add Unit


koay9f:
Add Unit to Volume
Imperial Gallon (imp gal)
1 imp gal = 153.772 fl-oz

Add Unit to Volumetric Flow Rate
Imperial Gallon per Minute (imp gpm)
1 imp gpm = 2.56203 fl-oz/s

Issue #781, opened 2017-10-18
Unit Converter - Gallons


koay9f:
Change name of "Gallons (gal)" to "U.S. Gallons (gal)"
(Nothing else changes)

PR #786, opened 2017-10-18
New Units for Unit Converter

RoundDev:
Issues #781 #783 

-  Gallons changed to U.S Gallons

-  New Volume unit Imperial Gallons

- New Volumetric Flow Rate unit Imperial Gallons per Minute


pshires:
I believe that this issue: https://github.com/npm/npm/issues/9696  is why the appveyor build failed here.

Issue #873, opened 2017-11-09
PHAST - Opening Losses Add Static Image to Help Text


pshires:
Need to add this graph to the help text for view factor / opening losses as a whole (maybe?)

pshires:
[PHAST_Lookup_ViewFactor.pdf](https://github.com/ORNL-AMO/AMO-Tools-Desktop/files/1459376/PHAST_Lookup_ViewFactor.pdf)

koay9f:
Could make it smaller in help text and if user clicks on it, makes a pop up with bigger one

AshaShibu:
Image broken on windows. Confirmed to work on Mac/linux

AshaShibu:
Image still broken on windows.
![capture8](https://user-images.githubusercontent.com/31292133/33679922-0c9da810-da8e-11e7-80c7-7cf878de96a5.PNG)

rmroot:
oops looks like the change in file extension did not save

Issue #872, opened 2017-11-08
PHAST non-EAF ET Energy Input


koay9f:
The result for "Energy Input" can be zero

If "Total Heat Input for burners" = zero, then the results "Fuel Heat Delivered" and "Exhaust Gas Losses" = zero, and that should be okay.

The form should be marked as completed so long as there is a number in "Total Heat Input for Burners"

PR #968, opened 2017-12-05
Issue #873 & #872

rmroot:
connects #873 
connects #872 
had .PNG instead of .png file extension


rmroot:
Also fixed a few small bugs in report

Issue #1118, opened 2018-01-10
PHAST- Add a new material -Liquid- Unit


AshaShibu:
The unit of Latent Heat of Vaporization should be changed to Btu/lb-F . Check unit conversion as well

koay9f:
![image](https://user-images.githubusercontent.com/25389736/34795506-92da1aa0-f620-11e7-877e-362986fcf6e5.png)

Latent Heat of Vaporization is not converting correctly.
I entered "32" for each field in imperial.  Latent Heat ... should = 133.977...

koay9f:
Opps.  okay, so the conversion is now correct (and it should be 32 Btu/lb, 74.432 kJ/kg), but the unit label is not.
It should be 
Btu/lb and kJ/kg

Again, JUST THE LABEL is wrong.  My fault.

AshaShibu:
It is now using the conversion for Btu/lb-F to kJ/kgC. It needs to be the conversion for Btu/lb to kJ/kg
The label is right but the conversion is now wrong

PR #1196, opened 2018-01-18
fix

rmroot:
connects #1118


src/app/phast/convert-phast.service.ts
```diff
@@ -191,6 +191,7 @@ export class ConvertPhastService {
   }
   //liquidChargeMaterial
   convertLiquidChargeMaterial(loss: LiquidChargeMaterial, oldSettings: Settings, newSettings: Settings): LiquidChargeMaterial {
+    debugger
```
pshires: just remove this then we'll be good. Kristina verified

Issue #1226, opened 2018-01-26
PH Calculator help text remove maxes


koay9f:
Some of these still have "Maximum Value Allowed" please remove all Maxes EXCEPT the following

O2 Enrichment:
O2 in combustion air
O2 in flue gasses (both)

Efficiency Improvement Calculator
Flue Gas Oxygen (both)

Energy Equivalency
ALL OF THESE ARE OKAY AS IS

Flow Calculations Energy Use
Coefficient of Discharge

Pre-Assessment
THESE ARE FINE

Issue #1225, opened 2018-01-26
O2 Enrichment Help Text


koay9f:
O2 in combustion Air
Has two help text fields.
Please remove 
see pic
![image](https://user-images.githubusercontent.com/25389736/35460254-5bfa55d4-02b1-11e8-8e0e-aae4ab1ae07e.png)

koay9f:
This looks like it might be a problem for multiple fields in this calculator

Issue #1224, opened 2018-01-26
Electrotechnology- EAF- Exhaust Gas - Help Text


AshaShibu:
Switch the description and input of Combustible Gas as CH4 with Total Volumetric Flow Rate and vice versa

Issue #1217, opened 2018-01-25
Typo- Help Text - Atmosphere Tab


AshaShibu:
Currently the help text title is Initial Temperature instead of  Inlet Temperature

PR #1259, opened 2018-02-06
Typos and Max Value Fixes

RoundDev:
connects #1217 
connects #1224 
connects #1225 
connects #1226 

- Fixing typo in help text atmosphere  Initial Temperature = Inlet Temperature

- Corrected help description for combustible Gas as CH4 and Total Volumetric Flow Rate

- Removed double help text for O2 in combustion Air

- Removed Maximum Values Allowed for calculators:

1. O2 Enrichment
2. Efficiency Improvement
3. Energy Equivalency
4. Flow Calculations Energy Use
5. Pre-Assessment


pshires:
The links to the calculators are broken when trying to access from the Process Heating tab, but work when accessed via All Calculators

Issue #847, opened 2017-11-02
Add images/graphs to the print view


gaccawi:
Make sure the graphs are added to the print view.

gaccawi:
For the graphs, keep in mind that printing may be done in black and white. One suggestion is to add the numbers (call-outs?)

PR #1275, opened 2018-02-09
Issue 847

DmitryHoward:
connects #847


rmroot:
Still needs some work before going into develop

Issue #1736, opened 2018-05-09
Fans - Additional Units


rmroot:

koay9f:
Figure out what units are for metric

koay9f:
Fluid
![image](https://user-images.githubusercontent.com/25389736/40940136-d740636e-6814-11e8-9d99-8f92e556f8ba.png)

Fan - No units
Motor - already good


Field Data
![image](https://user-images.githubusercontent.com/25389736/40940128-cfab83c2-6814-11e8-86df-6a0598903948.png)


Anything that was missed:
![image](https://user-images.githubusercontent.com/25389736/40940163-ee03defa-6814-11e8-816c-15ec3a5d16af.png)

koay9f:
In Metric, Flow Measurement needs a default (m3/hr)

PR #1869, opened 2018-06-08
issue-1736: reflected measurements in imperial and metric

FontenotZach:
connect #1736 
-We added conditional logic for displaying imperial/metric units of motor power, barometric pressure, static pressure, and gas density.  
-Changed "intel pressure" to "inlet pressure"
-Changed "mmHG" to "mmHg"


brappop:
We were unable to find mmh2o and m3/h to fix

PR #1969, opened 2018-06-25
fix small fsat issues

rmroot:
fixed settings bug.
fixed setup done bug.


rmroot:
fans branch has already been updated

Issue #1729, opened 2018-05-08
FSAT - Add Calculators


rmroot:
Setup fan calculators in assessment

rmroot:
Can setup and add AMCA 203 but may want additional calculators first?

koay9f:
Any motor calculators that are in PSAT can be added

koay9f:
System Curve:
Head --> Pressure, change units
Flow, change units
calculation should be okay

Achievable FAN Efficiency
Flow, change units
FAN type, use fan list.  for calculations use OptimalFanEfficiency.h I guess

Motor Perf. is good
NEMA Efficiency is good
Cut Specific Speed

FAN curve
Head --> Pressure, change units
Flow, change units
calculation should be okay

PR #2018, opened 2018-07-02
Issue 1729

rmroot:
connects #1729 

binding added for OptimalFanEfficiency

modified psat calcs to work with fans.

added calculator for optimal fan efficiency


rmroot:
Also merge latest from develop and fans-copy

Issue #2111, opened 2018-07-18
Additional Fan Icons


rmroot:
Missing Icons in report and export modals.

PR #2112, opened 2018-07-18
Issue 2111

rmroot:
connects #2111


rmroot:
PR also fixes scrolling bugs in report rollup

Issue #2330, opened 2018-09-10
PSAT (maybe FSAT) numbers changing without permission


koay9f:
unit conversions using input instead of copy?

don't know if FSAT is involved or not

koay9f:
More information about this bug:

The user changed all the numbers back to where they were supposed to be.  Ran a report: here is the user input data:

![image](https://user-images.githubusercontent.com/25389736/45318074-d785bf80-b509-11e8-8505-5e3bb05501f9.png)
Its all good.

Then he ran a rollup report including that pump.  here is the input data for the SAME pump
![image](https://user-images.githubusercontent.com/25389736/45318145-01d77d00-b50a-11e8-835d-ee321eb6e3f3.png)

no idea what is going on.

koay9f:
More info from user:

I have again been looking and it would appear to me that the Fluid temp, initially put in as 15 Deg C, it showing as 59Deg C, whereas 15DegC is in fact 59Deg F.
This would also be the same for flow (l/sec – USGPM)
This would be the same for head (m-ft)
The figures in the combined report seem to have been back calculated to your initial units, but are shown as my new units of l/sec, m & kW and from that it calculates the efficiency which is much higher than 100%.

----
I think it is actually doing this conversion every time it opens.  In the report above, temperature is at 999 deg C!

rmroot:
Should just be a PSAT problem

PR #2371, opened 2018-09-19
fix 2230

rmroot:
connects #2330


rmroot:
Should just be a PSAT problem

Issue #2448, opened 2018-10-12
FSAT add specified drive


koay9f:
add to drop down

DmitryHoward:
Input field is there, desktop data handling is implemented. No support on suite, so merging would break the app.

Issue #2447, opened 2018-10-12
PSAT/FSAT Explore opps drive


koay9f:
Add input for specified drive in explore ops

DmitryHoward:
Finished PSAT

DmitryHoward:
Finished FSAT

PR #2453, opened 2018-10-17
Issue 2447

DmitryHoward:
connects #2447 
connects #2448

### DIFF
--- a/src/app/fsat/compare.service.ts
+++ b/src/app/fsat/compare.service.ts
@@ -326,6 +326,23 @@ export class CompareService {
       modification = this.modifiedFSAT;
     }
     if (baseline && modification) {
+      if (baseline.fanSetup.specifiedDriveEfficiency != modification.fanSetup.specifiedDriveEfficiency) {
+        return true;
+      } else {
+        return false;
+      }
+    } else {
+      return false;
+    }
+  }
+  isSpecifiedDriveEfficiencyDifferent(baseline?: FSAT, modification?: FSAT) {
+    if (!baseline) {
+      baseline = this.baselineFSAT;
+    }
+    if (!modification) {
+      modification = this.modifiedFSAT;
+    }
+    if (baseline && modification) {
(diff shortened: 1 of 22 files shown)


src/app/fsat/compare.service.ts
```diff
@@ -319,6 +319,23 @@ export class CompareService {
     }
   }
   isDriveDifferent(baseline?: FSAT, modification?: FSAT) {
```
rmroot: specified-drive

I think these are backwords

src/app/fsat/explore-opportunities/explore-opportunities-form/fan-data-form/fan-data-form.component.ts
```diff
@@ -125,9 +127,21 @@ export class FanDataFormComponent implements OnInit {
     this.calculate();
   }
 
-  checkWarnings(){
+  checkWarnings() {
     this.specifiedError1 = this.fsatWarningService.checkFanWarnings(this.fsat.fanSetup).fanEfficiencyError;
     this.specifiedError2 = this.fsatWarningService.checkFanWarnings(this.fsat.modifications[this.exploreModIndex].fsat.fanSetup).fanEfficiencyError;
+    if (this.fsat.fanSetup.drive == 4) {
```
rmroot: Shouldn't need to add this logic here. This check is already being done in `fsatWarningsService` 

`this.specifiedError1` and `this.specifiedError2` will have `specifiedDriveEfficiencyError` member variable.

src/app/fsat/explore-opportunities/explore-opportunities-form/fan-data-form/fan-data-form.component.ts
```diff
@@ -146,6 +160,39 @@ export class FanDataFormComponent implements OnInit {
     }
   }
 
+  checkEfficiency(val: number, num: number) {
```
rmroot: This function is unneeded. See comment above. Warnings checked in warnings service.

Issue #2571, opened 2018-11-30
PSAT Motor Size warning


koay9f:
Rated Motor power is based on Field Data Motor power.  This value isn't in the modification.  Can it be compared to the result Motor Power (with unit conversion) instead?

![image](https://user-images.githubusercontent.com/25389736/49315501-e605c680-f4bb-11e8-9e7d-5fbc9a2fddb0.png)
I should be able to use a 25hp motor here...

PR #2582, opened 2018-12-04
changed comparison values for motor rated power

DmitryHoward:
connects #2571

### DIFF
--- a/src/app/fsat/fan-motor/fan-motor.component.ts
+++ b/src/app/fsat/fan-motor/fan-motor.component.ts
@@ -118,7 +118,14 @@ export class FanMotorComponent implements OnInit {
   }
 
   checkWarnings() {
-    this.warnings = this.fsatWarningService.checkMotorWarnings(this.fsat, this.settings);
+    let isMod: boolean;
+    if (this.fsat.modifications !== undefined && this.fsat.modifications !== null) {
+      isMod = false;
+    }
+    else {
+      isMod = true;
+    }
+    this.warnings = this.fsatWarningService.checkMotorWarnings(this.fsat, this.settings, isMod);
   }
 
   calcFla(): number {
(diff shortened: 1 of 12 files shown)


src/app/fsat/fan-motor/fan-motor.component.ts
```diff
@@ -118,7 +118,14 @@ export class FanMotorComponent implements OnInit {
   }
 
   checkWarnings() {
-    this.warnings = this.fsatWarningService.checkMotorWarnings(this.fsat, this.settings);
+    let isMod: boolean;
```
rmroot: Don't need this, `baseline` input is a poorly named boolean that is true when it is for the baseline and false when it is a modification. just use `!this.baseline`

src/app/fsat/fsat-tabs/fsat-tabs.component.ts
```diff
@@ -179,7 +179,14 @@ export class FsatTabsComponent implements OnInit {
     let fluidValid: boolean = this.fsatFluidService.isFanFluidValid(this.fsat.baseGasDensity);
     let fanValid: boolean = this.fanSetupService.isFanSetupValid(this.fsat.fanSetup, true);
     let motorValid: boolean = this.fanMotorService.isFanMotorValid(this.fsat.fanMotor);
-    let motorWarnings: FanMotorWarnings = this.fsatWarningService.checkMotorWarnings(this.fsat, this.settings);
+    let isMod: boolean;
```
rmroot: Don't need this. Fsat Tabs is always be for baseline. Modificaitons have their own tabs.

src/app/fsat/fsat-warning.service.ts
```diff
@@ -1,20 +1,28 @@
...
+import { FsatService } from './fsat.service';
 
 @Injectable()
 export class FsatWarningService {
 
-  constructor(private convertUnitsService: ConvertUnitsService, private psatService: PsatService) { }
+  constructor(private convertUnitsService: ConvertUnitsService, private psatService: PsatService, private fsatService: FsatService) { }
 
   checkFieldDataWarnings(fsat: FSAT, settings: Settings): FanFieldDataWarnings {
+    let isMod: boolean;
```
rmroot: have `isModification` be a passed in boolean instead of checking for modifications, doesn't always indicate baseline vs modification

Issue #2530, opened 2018-11-07
Unit Converter appearance


koay9f:
Can we have a limit on how wide the unit converter can be?  If I have the tool full screen, it stretches across the screen and looks bad.

PR #2651, opened 2019-01-17
Fixed size of Unit Converter so it looks cleaner

Aeledbetter:
*Fixed size of Unit Converter so it looks cleaner
connects #2530


Aeledbetter:
Done

Aeledbetter:
Looks good.

Issue #2532, opened 2018-11-07
Steam Turbines & Explore Opps


koay9f:
Have possible actions hidden if not applicable 

If only 1 header then only condensing turbine

If only 2 header then only condensing turbine & high to low turbine

koay9f:
What branch is this on?  I just looked at steam, with a 1 header system and all the "possible" turbines are still showing up

koay9f:
2/1/19 - Made 1 header assessment, HP to MP & MP to LP turbines still in the list of "potential Adjustment Projects"

koay9f:
1 have a one header system, the only turbine you should see on the Explore Opportunities page is "Condensing Steam Turbine"
![image](https://user-images.githubusercontent.com/25389736/52137629-04518500-2619-11e9-8819-7958054eeb2d.png)

If it is a two header system, then you should only see "Condensing Steam Turbine" and "High Pressure to Low Pressure Steam Turbine"
![image](https://user-images.githubusercontent.com/25389736/52137694-2a772500-2619-11e9-963c-85cc8f97409a.png)

PR #2674, opened 2019-01-25
Issue 2532

DmitryHoward:
connects #2532


rmroot:
This should be merged into steam not develop

Issue #2682, opened 2019-01-25
PSAT Rename


koay9f:
When you use the backspace to delete the Modification Name, `Opportunities Modification` pops up, if you use backspace to delete that, it returns...
only happens in PSAT

![rename with backspace](https://user-images.githubusercontent.com/25389736/51770297-5fccc180-20b3-11e9-91fe-74058b94ff51.gif)

PR #2691, opened 2019-01-28
Possible fix for issue-2682

Aeledbetter:
connects #2682 

Please check to confirm this is fixed, Looks like this code was causing the text to return when backspaced:
``if (!this.psat.modifications[this.modificationIndex].psat.name) {
      this.psat.modifications[this.modificationIndex].psat.name = 'Opportunities Modification';
    }``
Also, removing this code didn't seem to break anything, as it wasn't in the code for the other assessments, which seems to be why it wasn't happening there too.
 
I was able to stop the bug from happening when a new assessment is added. When the default "Scenerio1" text is deleted via backspace, "Opportunities Modification" no longer appears, also tested with adding other scenarios as well.

### DIFF
--- a/src/app/phast/explore-phast-opportunities/explore-phast-opportunities.component.ts
+++ b/src/app/phast/explore-phast-opportunities/explore-phast-opportunities.component.ts
@@ -44,7 +44,6 @@ export class ExplorePhastOpportunitiesComponent implements OnInit {
   modExists: boolean = false;
   selectModificationSubscription: Subscription;
   toastId: any;
-  emitSave: any;
   constructor(private lossesService: LossesService, private toastyService: ToastyService,
     private toastyConfig: ToastyConfig,
   ) {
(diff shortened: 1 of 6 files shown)


src/app/phast/explore-phast-opportunities/explore-phast-opportunities.component.ts
```diff
@@ -59,7 +60,11 @@ export class ExplorePhastOpportunitiesComponent implements OnInit {
   ngOnDestroy() {
     this.toastyService.clearAll();
     this.exploreOppsToast.emit(false);
-  }
+      if(this.phast.modifications[this.exploreModIndex] && !this.phast.modifications[this.exploreModIndex].phast.name){
+        this.phast.modifications[this.exploreModIndex].phast.name = 'Opportunities Modification';
+        this.emitSave.emit(true);
```
rmroot: You should use the "save" event emitter declared as an Output. This emit save won't do anything.

Issue #3006, opened 2019-05-10
TH Opp Sheet Tweeks


koay9f:
Change titles for top fields

Project Title (top left)
Project Owner/Lead (top right)
Process/Equipment (bottom left)
Plant/Business Unit  (bottom right)

shubhamkokul:
Which section holds these titles and what should be the new names.

PR #3025, opened 2019-05-17
Update field titles in Opp sheet

shubhamkokul:
connects #3006


CMakeCache.txt
```diff
@@ -0,0 +1,38 @@
+# This is the CMakeCache file.
```
DmitryHoward: This file should not be included in the commit. Please remove.

CMakeFiles/cmake.check_cache
```diff
@@ -0,0 +1 @@
+# This file is generated by cmake for dependency checking of the CMakeCache.txt file
```
DmitryHoward: This file should not be included in the commit. Please remove.

Issue #3041, opened 2019-05-21
Compressed Air Reduction Help Text


DmitryHoward:
Add help text for compressed air reduction calculator.

koay9f:
For the help text that always appears in the calculator, the list (turning off compressed air..., reducing compressed air..., eliminating inappropriate..., using engineered nozzles...) should be bulleted.  Don't think it needs lines between each...

Add field specific help text for "Utility Type"
"Choose how compressed air is delivered to your facility.  Choosing "Compressed Air" means it enters your facility as compressed air; choosing "electricity" means it is compressed on site."

Issue #2822, opened 2019-03-01
Compressed Air Reduction TH Calculator


vanced:

vanced:
[Compressed Air TH help and math Final.docx](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5b18368716b0dc00135e3482/125516c66e82c728ace21e0d46cb9b97339086ba9cecbb038554e2d41d6e677b2c08b36b718be768b2436f094f2917f210044d58eeb9cb74fce36f34985619e8d8390a26973f45a5bfd349ed53ba208e48d3b7b9d91ec6fa2137a5dad11b61435c52dcbd4f3fb29843c259f34ee04953a2103e6ae37e346e0a.docx)
![Compressed Air Reduction TH Figure.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5b18368716b0dc00135e3482/125516c66e82c728ace21e0d46cb9b97339086ba9cecbb038554e2d41b4323662a10aa247ec5d700ff6472061a1d1bbe5d4a1f06a3a3dc78e2a436628f555fb8833c5d20cf3a40faead01fbf04bd768a4ad6e4ba9850d2aa6c7bb3d4d216744b470a83ed5c32aa984ecd59f24de74e5ca2113062f43b.png)

koay9f:
weeks/year
days/week
shifts/day
hours/shift
--> hours/year

koay9f:
UI Update.  New field "Utility Type" will determine if the electricity section shows up, utility cost units, and what utility the opp sheet will be

![image](https://user-images.githubusercontent.com/25389736/57151091-1240ae80-6d9e-11e9-90f7-7dd29958a4b1.png)

koay9f:
For orifice method, have the "Nozzle Flow Rate" result shown before "Number of Nozzle" input

koay9f:
**Left side: Inputs**
*Utility Type Dropdown*
* Change "None" to "Compressed Air"
* Should be locked for whole calculator.  (only choose once)

*Need cost fields.  either...*
1 - one that's name changes and what is multiplies to changes (if utility = compressed air, then it is compressed air cost and annual cost = unit cost * consumption; if utility = electricity, then it is electricity cost and annual cost = unit cost * energy use)
[...]

PR #3053, opened 2019-05-23
Compressed air reduction

DmitryHoward:
connects #2822 
connects #3041


rmroot:
Need to fix package.json AMO-Tools-Suite version

Issue #3022, opened 2019-05-16
TH - Treasure chest design updates


rmroot:
Updates treasure chest to support multiple gauges for Opp sheets with more then one utility type

rmroot:
Make sure correct utilities are listed in summary card.

Do we want to add additional utility gauges for overall summary?

PR #3092, opened 2019-06-04
Issue 3022

rmroot:
connects #3022


src/app/treasure-hunt/find-treasure/find-treasure.component.html
```diff
@@ -103,9 +90,15 @@ <h1 class="p-2 lead text-center mb-0">
       </div>
     </div>
 
+<<<<<<< HEAD
```
DmitryHoward: git merge header still present in html code

Issue #2937, opened 2019-04-24
Fan Analysis Needs Work


koay9f:
Notes from Tom:

Input Plane Data
* Have the "Add Differential Pressure Readings" not be a separate page, just put the table below "Input Velocity Pressure Traverse Data"
* Name the ^ section "Pitot Tube Traverse Readings"
* Show result for Pv3 (label = "Calculated Velocity Pressure")
* any ways to make that data entry easier?
-             - copy/paste from Excel?
-              - buttons to "duplicate across row" or column
* Make it so don't have to finish "Fan Shaft Power" to get some results.  Make it like how do we do it in Fan Assessment for "Calculate Pressures and Flow Rate", can get "Plane Data Table" as soon as plane 5 data entry is complete.
* Next button for moving between planes
* BIG ASK - when on a plane, have something highlighting the diagram showing where you are.  Would this require pictures for each plane?  Could we color code the planes on the pic and have the navigation buttons match?

Fan Shaft Power
* BUG: Estimate Full Load Amps doesn't Work
* remove the `(Hi)` from `Motor Input Power (Hi)` (both the result and the entry field)
![image](https://user-images.githubusercontent.com/25389736/56676897-4543bd80-668d-11e9-97a6-d806d62d4d85.png)

Overall
* Navigation is difficult - Instead of having to "Return to Setup" could we also have "Next"
* Move Reset Data to the left side
* BUG: Results do not show up.  I think this is a problem with "Establish Base Gas Density".  It will work if you toggle "Method to Establish Gas Density"
Errors:
![image](https://user-images.githubusercontent.com/25389736/56676677-d7979180-668c-11e9-807d-ad1af8566d32.png)

koay9f:
Re: Navigation - in addition to "next" maybe have similar layout to assessments w/ 1, 2, 3, 4, Results, in a top bar?

No results for Basic Fan Info, Establish Gas Density
Once `Make it so don't have to finish "Fan Shaft Power" to get some results` goes through, Input Plane Data should have results

Have Next for navigation in `Input Plane Data`

koay9f:
The overall change in navigation is much better.  A few things missing/updated from original issue text.


[3] Plane Data
* Instead of "Name the ^ section "Pitot Tube Traverse Readings"" maybe change name [3] Plane Data to [3] Traverse Data    (its a lot shorter)
[...]

Issue #3129, opened 2019-06-12
Calculator name and description overhaul


koay9f:
Calc descriptions (in the calc lists, not the ones in the calc help text) are often too long or not very descriptive of the purpose of the calc.
Names don't always match between the various places they are displayed 

Revision of description:
[https://app.zenhub.com/files/80439269/577e5041-bec5-4dfe-9d02-64798c548d3e/download](https://app.zenhub.com/files/80439269/577e5041-bec5-4dfe-9d02-64798c548d3e/download)

koay9f:
Some of this may be impacted by the quick fixes.  Let me know if any conflicts occur

PR #3214, opened 2019-07-24
Issue 2937

rmroot:
connects #2937 

Fan analysis overhaul. 

Additional issues added for removing fan203 component completely and updating it's use in Fan Assessments. 

Also added issue for easier reading data entry.


rmroot:
connects #3129

Issue #2820, opened 2019-03-01
Steam TH Calculator


vanced:

vanced:
[Steam TH help and math Final.docx](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5b18368716b0dc00135e3482/125516c66e82c728ace21e0d46db809f228fc39da7a9b746a84db0952742677e2810ab6b568ced29b30037431f1d11f6140b1b56f6eb8c22f2e36e21880802ea803b0d26c63d14abee831aec53bc32975dc7e7e2de139db17770a599855d3f190f198cfa5c31a69c43cc50f44be15f18.docx)

vanced:
![Steam TH Figure.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5b18368716b0dc00135e3482/125516c66e82c728ace21e0d46db809f228fc39da7a9994aa348e2916b0a65633b0ba92e7391a172fd1722050a0d1fae15515815aaabdb2ba6b33b268b5958bfd77f4366846c4ff8e7df46f554f832815d84b1b68b49c5e93923e2828f483b1e020ac4.png)

koay9f:
Operating hours should be calculated like the rest of the tool
weeks/year
days/week
shifts/day
hours/shift
--> hours/year

koay9f:
Add dropdown near "Fuel Cost" for "Fuel Type". 
Options include: "Natural Gas" (Default), "Steam as Utility", "Other Fuel"

koay9f:
UI Revamp.  should take in to account the above comments.  Open to some reorganization of fields
![image](https://user-images.githubusercontent.com/25389736/57143673-5e372780-6d8d-11e9-8cb3-7f0814ba1f8e.png)

koay9f:
* Utility Type should be locked for the calculator, like Compressed Air.  Choose once and then it is that for the other equipment and modification

* Pressure is not being converted to MPaa before being sent to suite

* Pressure default value should be 100 psig (which is 0.790801 MPaa, what the default value currently is...)

* Flow Rate units label should be lb/hr (imperial) and kg/hr (metric)

* Flow Meter Method
-- Steam use result is not correct (also putting in Math issue, but the math in the suite looks okay so I'm not sure what the problem is)
[...]

PR #3258, opened 2019-08-27
Issue 2820

DmitryHoward:
connects #2820


rmroot:
@DmitryHoward package.json has a local tools suite path. That's why build error in CI here.

rmroot:
steam-mod-disable

input fields for modification are not disabling when not focused. (baseline is disabling)

Issue #3363, opened 2019-10-10
Badges colors


koay9f:
Steam: The number turns red on error instead of yellow like the rest of the tool

Pumps: 
When you create a new scenario, Pump & Fluid is blue (different from BL).  I'm guessing b/c of "Pump Type" being "specified" in the background.  (Especially since if you change it to "optimized" then it turns back to green, which is also bad)
Thoughts:
Can we make it so that when it checks for differences, it ignores `Pump Type` and instead checks if the `Pump Efficiency` is equal to that calculated by BL?  Otherwise we need to brainstorm.

Fans: 
Fan is always blue. Can't make it turn green.  Same solution as pumps?

koay9f:
Pumps: 
* hover over the badges doesn't produce a text box (while playing in fans I was able to get the hover to come up but it took a long time, okay like 2 seconds, but still.  also have to be really exact)
* the pump efficiency compairson worked at first (showed no difference), but if you change it once, then you can never get it to go back to green and new scenarios also show a difference.  Maybe have it compare after rounding to a single decimal point?

Fans: same as pumps

Steam: ... do we have any yellow level errors?  do all the warnings in steam stop calculations?  - message me when you look at this again.

koay9f:
Pumps: using "optimize pump" does not change the badge.  Could we have the "optimal pump efficiency" replace the value in the now hidden pump efficiency field?  Not sure.  But if using optimized efficiency should then be blue badge.  

Fan: using "optimize pump" does not change the badge.  Also, default fan efficiency (based on baseline) turns the badge blue.  Maybe rounding problem like was having with pumps?

PR #3384, opened 2019-10-23
Fix badge colors

DmitryHoward:
connects #3363


src/app/shared/tabs-tooltip/tabs-tooltip.component.ts
```diff
@@ -21,7 +21,7 @@ export class TabsTooltipComponent implements OnInit {
     this.setMessage();
   }
 
-  ngOnChanges(changes: SimpleChanges) {
+  ngOnChanges(changes: SimpleChanges) {1
```
rmroot: Dont think that 1 is supposed to be there.

Issue #3582, opened 2020-02-20
PSAT Sankey 2.0


rmroot:
Change psat sankey to plotly.js usage

nbintertech:
Current chart below. Estimating 3 points left for additional styling tweaks, testing and development time to merge in with changes since it was last current with develop branch.
![psat.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/92fc9b65-3832-4674-8e19-2ec63f19d49c)

koay9f:
Two problems:

In Novice View - Sankey is not updating as I change things, I have to switch between BL and mod to refresh the page

If I switch to direct drive, then I get the "greater than 100%" message.  I'm guessing this is b/c drive efficiency is now at 100%...?  100% is fine, greater than 100% is bad.

PR #3607, opened 2020-02-28
Refactor psat-sankey to Plotly

nbintertech:
connects #3582 

- Creates shared psat-sankey module/component that uses Plotly.js
- Removes old psat/sankey
- Removes explore-opportunities-sankey and wrapping component


src/app/shared/psat-sankey/psat-sankey.component.ts
```diff
@@ -307,15 +305,38 @@ export class PsatSankeyComponent implements OnInit {
...
     return nodes;
   }
 
+  buildLinks(nodes, links) {
+    this.connectingLinkPaths.push(0);
+    
+    for (let i = 0; i < nodes.length; i++) {
+      if (nodes[i].isConnector) {
+          this.connectingNodes.push(i); 
+          if (i !== 0 && i-1 !== 0) {
```
nbintertech: These connectingLinkPaths may be better off set from a isConnectingLink property on the node.

PR #3630, opened 2020-03-09
Issue 3505

causeyc:
Refactored Travis CI, moved all if blocks to separate script files in order to allow Travis to recognize that the build has failed if any command returns a nonzero exit status.


gaccawi:
Approved

Issue #3637, opened 2020-03-11
Show Mods in Report


nbintertech:
Child issue of #3635

Selected 'savings opportunities'/ modifications as well as 'energy projects' should be reflected in report > result data

nbintertech:
@koay9f I've verified that the checkbox for each modification operates as intended when moving between assessment and report, but please double-check me on this during verification.

PR #3643, opened 2020-03-13
Show current assessment modifications in fan report

nbintertech:
connects #3637 

- Fans report reflects changes made to both modifications (fan, motor, etc) and energy projects within the assessment


src/app/fsat/explore-opportunities/explore-opportunities-form/explore-opportunities-form.component.ts
```diff
@@ -110,6 +110,9 @@ export class ExploreOpportunitiesFormComponent implements OnInit {
   }
 
   initForms() {
+    if (!this.fsat.modifications[this.exploreModIndex].exploreOppsShowVfd) {
```
rmroot: old assessments will have an fsat.isVFD value that should be used for the `hasOpportunity: false`

src/app/fsat/fsat-report/results-summary/results-summary.component.css
```diff
@@ -15,6 +15,10 @@ tr > th:not(:first-child){
     text-align: center;
 }
 
+.generated-cell-item {
```
rmroot: bootstrap has built in class of `d-block` for display: block

Issue #3644, opened 2020-03-13
Explore-Opps Fans - Disabled/Selected baseline value


nbintertech:
Child issue of #3625

FSAT > Assessment > Explore-Opportunities
The user would like to see the baseline and modificaton values side by side

PR #3649, opened 2020-03-16
(Fans) Enhance explore-opportunities baseline and mod UI

nbintertech:
connects #3644 

- Fan UI enhancement to place explore-opportunities baseline and modification selects side by side

### DIFF
--- a/src/app/fsat/explore-opportunities/explore-opportunities-form/system-data-form/system-data-form.component.html
+++ b/src/app/fsat/explore-opportunities/explore-opportunities-form/system-data-form/system-data-form.component.html
@@ -5,28 +5,16 @@
         Reduce System Flow Rate
     </label>
 </div>
-<li *ngIf="showFlowRate && !isVFD">
-    <form [formGroup]="baselineForm">
-        <div class="form-group" [ngClass]="{'invalid': baselineForm.controls.flowRate.invalid}">
-            <label for="baselineFlowRate">Baseline Flow Rate</label>
-            <div class="input-group">
-                <input name="baselineFlowRate" type="number" step="any" id="baselineFlowRate" class="form-control"
-                    (focus)="focusField('flowRate')" (input)="calculate()" formControlName="flowRate" />
-                <span class="input-group-addon units" [innerHTML]="settings.fanFlowRate | settingsLabel"></span>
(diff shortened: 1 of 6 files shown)


src/app/fsat/explore-opportunities/explore-opportunities-form/system-data-form/system-data-form.component.html
```diff
@@ -5,28 +5,16 @@
...
-                <span *ngIf="baselineForm.controls.flowRate.errors.required">Value Required</span>
-                <span *ngIf="baselineForm.controls.flowRate.errors.min">Value can't be less than
-                    {{baselineForm.controls.flowRate.errors.min.min}} <span
-                        [innerHTML]="settings.fanFlowRate | settingsLabel"></span></span>
-            </span>
+<li *ngIf="showFlowRate && !isVFD" class="explore-opps-item">
+    <form [formGroup]="baselineForm" class="explore-opps-form">
+        <div class="form-group explore-opps-formgroup" [ngClass]="{'invalid': baselineForm.controls.flowRate.invalid}">
+            <label for="baselineFlowRate" class="baseline-value-label">Baseline Flow Rate</label>
+            <div class="text-muted baseline-value-input">{{ baselineForm.controls.flowRate.value }} {{settings.fanFlowRate || settingsLabel }}</div>
```
rmroot: Let's add the number pipe onto this.

Additionally, pipes only use one | so it should be `| settingsLabel` we also use [innerHTML] for the settingsLabel because it has hex codes for super and subscripts

rmroot: Apply in all places using pipes

src/assets/styles/explore-opps.css
```diff
@@ -0,0 +1,64 @@
...
+    padding-bottom: 4px;
+    align-self: center;
+  }
+  
+  div.form-group.explore-opps-formgroup >
+  select.mod-value-input {
+    width: 100%;
+  }
+
+  hr.form-break-rule {
```
rmroot: probably don't need to make custom class here, bootstrap has `mt-0` for `margin-top: 0`, `w-100` for `width: 100%`

Issue #3656, opened 2020-03-20
(Pumps) Show Mods in Report


nbintertech:
Child issue of #3635

Selected 'savings opportunities'/ modifications as well as 'energy projects' should be reflected in report > result data

PR #3659, opened 2020-03-23
Reflect PSAT selected opps in report

nbintertech:
connects #3656 

- Show assessment selected energy projects and mods in PSAT report
- Remove 'this' refs from FSAT templates

### DIFF
--- a/src/app/fsat/explore-opportunities/explore-opportunities-form/explore-opportunities-form.component.ts
+++ b/src/app/fsat/explore-opportunities/explore-opportunities-form/explore-opportunities-form.component.ts
@@ -110,7 +110,10 @@ export class ExploreOpportunitiesFormComponent implements OnInit {
   }
 
   initForms() {
-    if (!this.fsat.isVFD && !this.fsat.modifications[this.exploreModIndex].exploreOppsShowVfd) {
+    if (this.fsat.modifications[this.exploreModIndex].fsat.isVFD) {
+      this.fsat.modifications[this.exploreModIndex].exploreOppsShowVfd = {hasOpportunity: true, display: "Install VFD"};
+      delete this.fsat.modifications[this.exploreModIndex].fsat.isVFD;
+    } else if (!this.fsat.modifications[this.exploreModIndex].exploreOppsShowVfd) {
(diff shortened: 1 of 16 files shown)


src/app/fsat/explore-opportunities/explore-opportunities-form/fan-data-form/fan-data-form.component.html
```diff
@@ -127,7 +127,7 @@
     </div>
   </form>
 </li>
-<li *ngIf="showFanType || isVFD">
+<li *ngIf="fsat.modifications[exploreModIndex].exploreOppsShowFanType || isVFD">
```
rmroot: I think this should be `.hasOpportunity` otherwise it will be showing if hasOpportunity is true or false

src/app/fsat/explore-opportunities/explore-opportunities-form/explore-opportunities-form.component.ts
```diff
@@ -110,7 +110,7 @@ export class ExploreOpportunitiesFormComponent implements OnInit {
   }
 
   initForms() {
-    if (!this.fsat.isVFD || !this.fsat.modifications[this.exploreModIndex].exploreOppsShowVfd) {
+    if (!this.fsat.isVFD && !this.fsat.modifications[this.exploreModIndex].exploreOppsShowVfd) {
```
rmroot: I don't think this is the correct logic.

Use an installed version of the app to select isVFD and then close the app and run this branch. The isVFD isn't selected.

src/app/psat/explore-opportunities/explore-opportunities-form/explore-opportunities-form.component.ts
```diff
@@ -82,6 +82,9 @@ export class ExploreOpportunitiesFormComponent implements OnInit {
   }
 
   initForms() {
+    if (!this.psat.inputs.isVFD && !this.psat.modifications[this.exploreModIndex].exploreOppsShowVfd) {
```
rmroot: Same as the comment from fsat. Test this with and installed version, select is vfd and then run this version. isVFD should be selected but I don't think it will be. This is probably the problem area.

Issue #3682, opened 2020-04-03
Invalid indication for phast mods


rmroot:
Add error message to phast reports when modification is invalid and indicate which portion of modification is invalid

rmroot:
- If baseline is invalid disable all tabs in the banner except setup, if you make the baseline invalid in the "Assessment" tab you should be able to stay in that tab but if you leave the entire assessment it should bring you back to setup the next time you go to that phast assessement.


There needs to be an indication that there is an invalid setup for modifications:

- Executive Summary
- Result Data
- Report Graphs (don't show, like psat/fsat)
- Sankey (don't show, like psat/fsat) report and tab
- Input Summary
- Also need to indicate invalid setup in the results tab of the assessment

This is done in fsat/psat already, use a similar messaging style. Logic probably be a little different. I added FsatValid and PsatVaild interfaces for those, will most likely have to do something similar for phast

PR #3700, opened 2020-04-15
Issue 3682

nbintertech:
connects #3682 

- Add invalid forms messaging for modification results, and reports areas
- Minimal refactor of heat-system-efficiency to get desired behavior (but seems like it could be refactored further)

### DIFF
--- a/src/app/phast/phast-report/executive-summary/executive-summary.component.html
+++ b/src/app/phast/phast-report/executive-summary/executive-summary.component.html
@@ -13,16 +13,14 @@
         <td class="percent-savings">Percent Savings (%)</td>
         <td [ngClass]="{'selected-modification': selectedModificationIndex == -1}">&mdash; &mdash;</td>
         <td *ngFor="let mod of modifications; let index = index;" [ngClass]="{'selected-modification': index == selectedModificationIndex}">
-
-          <div class="hide-print" *ngIf="mod.percentSavings">
-            
+          <app-invalid-phast *ngIf="!phast.modifications[index].phast.valid.isValid" [phastValid]="phast.modifications[index].phast.valid"></app-invalid-phast>
+          <div class="hide-print" *ngIf="mod.percentSavings && phast.modifications[index].phast.valid.isValid">
               <app-percent-graph [value]="mod.percentSavings"></app-percent-graph>
-            
           </div>
-          <div class="show-print text-center" *ngIf="mod.percentSavings">
+          <div class="show-print text-center" *ngIf="mod.percentSavings && phast.modifications[index].phast.valid.isValid">
             {{mod.percentSavings | number:'1.0'}} %
           </div>
-          <div *ngIf="!mod.percentSavings">
+          <div *ngIf="!mod.percentSavings && phast.modifications[index].phast.valid.isValid">
             &mdash; &mdash;
           </div>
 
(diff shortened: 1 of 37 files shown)


src/app/phast/phast-report/executive-summary/executive-summary.component.html
```diff
@@ -13,16 +13,33 @@ <h4 class="show-print print-section-header">Assessment Executive Summary</h4>
         <td class="percent-savings">Percent Savings (%)</td>
         <td [ngClass]="{'selected-modification': selectedModificationIndex == -1}">&mdash; &mdash;</td>
         <td *ngFor="let mod of modifications; let index = index;" [ngClass]="{'selected-modification': index == selectedModificationIndex}">
-
-          <div class="hide-print" *ngIf="mod.percentSavings">
-            
+          <div class="alert-danger" *ngIf="!phast.modifications[index].phast.valid.isValid">
```
rmroot, line 16: In here you should be able to use the "mod" object from the ngFor... i **think**

src/app/phast/phast-report/report-graphs/report-graphs.component.html
```diff
@@ -7,9 +7,29 @@
       </select>
     </div>
     <div class="p-2 pie-chart-container">
-      <app-plotly-pie-chart [valuesAndLabels]="selectedBaselineData.valuesAndLabels" [valuesUnit]="lossUnit">
+      <app-plotly-pie-chart *ngIf="selectedBaselineData.valid.isValid" [valuesAndLabels]="selectedBaselineData.valuesAndLabels" [valuesUnit]="lossUnit">
       </app-plotly-pie-chart>
     </div>
+    <div class="alert-danger small" *ngIf="!selectedBaselineData.valid.isValid">
```
rmroot, line 13: This appears to be a repeated logic as other places. Should make a shared component in phast that takes the valid object and does this display logic. Didn't do that in psat/fsat because it only had 4 lines

src/app/phast/phast.service.ts
```diff
@@ -891,5 +907,291 @@ export class PhastService {
   //   })
   //   return sum;
   // }
+
+
```
rmroot, line 911: Create a new service for this validity check please

src/app/phast/phast.service.ts
```diff
@@ -891,5 +907,291 @@ export class PhastService {
...
+        if (otherLossForm.status === 'INVALID') {
+          valid = false;
+        }
+      });
+    }
+    return valid;
+  }
+
+  checkSysEfficiencyValid(phast: PHAST): boolean {
+    const efficiencyValue = this.heatSystemEfficiencyCompareService.modification.systemEfficiency;
```
rmroot, line 1166: "let"

should be `phast.systemEfficiency` 

using compare service will only ever check the selected modification which is not the desired behavior

rmroot: Others look good, guessing copy/paste thing

src/app/phast/phast-report/executive-summary/executive-summary.component.html
nbintertech: That mod is a summary of results. If I keep as is, I won't have to modify the results array by adding 'valid' and assigning. Not sure which is preferable though.

Issue #3681, opened 2020-04-03
Invalid indication for steam mods


rmroot:
Add warning that steam modification has an invalid setup in the report as well as the tab that is invalid

PR #3713, opened 2020-04-20
Issue 3681

nbintertech:
connects #3681 

- Add invalid messaging for SSMT modifications

### DIFF
--- a/src/app/ssmt/ssmt-report/report-graphs/ssmt-pie-chart/ssmt-pie-chart.component.html
+++ b/src/app/ssmt/ssmt-report/report-graphs/ssmt-pie-chart/ssmt-pie-chart.component.html
@@ -1,5 +1,5 @@
 <div #ssmtPieChart [ngStyle]="{'display': ssmt.valid.isValid? 'block' : 'none'}"></div>
-<div class="w-100 text-center" *ngIf="noData">
+<div class="w-100 text-center" *ngIf="noData && ssmt.valid.isValid">
     No Data for Scenario
 </div>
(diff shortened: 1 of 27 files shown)


src/app/ssmt/ssmt-report/report-graphs/ssmt-pie-chart/ssmt-pie-chart.component.html
```diff
@@ -1,4 +1,11 @@
-<div #ssmtPieChart></div>
+<div #ssmtPieChart [ngStyle]="{'display': ssmt.valid.isValid? 'block' : 'none'}"></div>
 <div class="w-100 text-center" *ngIf="noData">
```
rmroot, line 2: Our check for noData is if the turbines generate energy. They are showing up for invalid as well. Should only show up if valid and no generation

src/app/ssmt/ssmt-report/ssmt-report.component.ts
```diff
@@ -157,11 +159,20 @@ export class SsmtReportComponent implements OnInit {
 
   calculateResultsWithMarginalCosts(ssmt: SSMT, outputData: SSMTOutput, baselineResults?: SSMTOutput): SSMTOutput {
     let marginalCosts: { marginalHPCost: number, marginalMPCost: number, marginalLPCost: number };
+
     if (ssmt.name == 'Baseline') {
       marginalCosts = this.ssmtService.calculateBaselineMarginalCosts(ssmt, outputData, this.settings);
+    } else if (!ssmt.valid.isValid) {
+        outputData = baselineResults;
```
rmroot, line 166: This is setting the modification output to the baseline results, we don't want to do that since it's the incorrect results.

Shouldn't have this else if statement..

Issue #3764, opened 2020-06-01
fan compressibility factor in mod


koay9f:
it needs the little link/button to calc compressiblity factor in expert view

it first needs the calcs to work.
stop gap for now:
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/8da2ba64-c73e-4e16-ba6b-92c019e3edb4)

koay9f:
![compressFactorBug.gif](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/f1c7e40c-4b20-429e-84fa-379a1ef10ee9)


Compressibility Factor doesn't calculate right away.  You have to click define, set a value, then click calculate again.

Math itself looks good

PR #3800, opened 2020-06-15
Issue 3764

nbintertech:
connects #3764 

Adds compressibility factor calculation for mods.


src/app/fsat/fan-field-data/fan-field-data.component.html
```diff
@@ -184,8 +184,8 @@ <h3>Field Data</h3>
     <div class="form-group"
       [ngClass]="{'indicate-different': isCompressibilityFactorDifferent(), 'invalid': fieldDataForm.controls.compressibilityFactor.invalid}">
       <label for="{{'compressibilityFactor_'+idString}}">Compressibility Factor
-        <a class="form-text small click-link" *ngIf="baseline" (click)="calculateCompressibility()"
-          [ngClass]="{'disabled-link':!selected}">Calculate
```
rmroot, line 188: We still want this disable when not selected

src/app/fsat/fan-field-data/fan-field-data.component.ts
```diff
@@ -162,21 +162,43 @@ export class FanFieldDataComponent implements OnInit {
...
       barometricPressure: this.fsat.baseGasDensity.barometricPressure,
       flowRate: this.fieldDataForm.controls.flowRate.value,
       specificHeatRatio: this.fieldDataForm.controls.specificHeatRatio.value
     };
-    let calcCompFactor: number = this.fsatService.compressibilityFactor(inputs, this.settings);
+
+    if (this.baseline) {
+      compressibilityFactor = this.fsatService.compressibilityFactor(inputs, this.settings);
+    } else {
+      let currentMoverShaftPower;
```
rmroot, line 181: Can you more this into a function that passes the parameters and returns the calculated value.

Issue #3778, opened 2020-06-09
DataExplorer: Save log tool state thru close


rmroot:
Add log tool to indexed db so that data can be stored/accessed later

PR #3805, opened 2020-06-17
Issue 3778

rmroot:
connects #3778 

log tool data saving in indexedDb and logic setup for using on next session.

indexeddb version 5


src/app/log-tool/log-tool-db.service.ts
```diff
@@ -0,0 +1,137 @@
...
+          this.logToolDbData = logToolDbData;
+          this.previousDataAvailable.next(logToolDbData[0].modifiedDate);
+        }
+      })
+    }
+  }
+
+  addLogToolToDb() {
+    let logToolDbData: LogToolDbData = {
+      // id: undefined,
```
nbintertech, line 39: A few other comments to review if needed.

Issue #3827, opened 2020-07-07
Utility cost in Compressed air leak survey


koay9f:
If you type 0.0 it automatically removes the decimal place so the next number you enter is an integer.
You can type 0.X and then put in a number between the 0. and the X (0.0X)

rmroot:
**### VERIFY generate example and reset buttons are working properly**

PR #3828, opened 2020-07-07
fix utility cost form input in compressed air leak survey

rmroot:
connects #3827


rmroot:
FYI: I auto-formatted the html file but it didn't have any real changes.

Issue #3806, opened 2020-06-18
Plotly Update Pump Curve


nbintertech:
Child of 3757

Update Pumps > Pump Curve to Plotly

PR #3851, opened 2020-07-14
Issue 3806

nbintertech:
connects #3806


nbintertech:
Moving work to system-curve branch for epic

Issue #3848, opened 2020-07-10
Motor Inventory: Tie to dashboard folders


rmroot:
Integrate motor inventory module into dashboard folders for saving

PR #3929, opened 2020-08-18
Issue 3848

rmroot:
connects #3848 

added saving inventories to folders
tied settings to inventory items
import/export inventories


rmroot:
Merged latest from develop.. jumped to 298 files..

Issue #3849, opened 2020-07-10
Update NPM packages


rmroot:
Check for outdated npm packages...

Know forsure electron is behind.

PR #3995, opened 2020-09-22
Issue 3849

rmroot:
connects #3849 

fixes prod build bugs

electron v4 -> 10
angular v9 -> 10

additional changes to deal with angular and electron deprecation's from updates

updated readme to use update scripts


rmroot:
Going to fail CI until tools suite v0.8.0 is released

Issue #4064, opened 2020-10-20
Steam Fix - Loss page 


koay9f:
Blowdown losses are not calculated correctly if we have a blowdown flash tank.  If that is the case then...
Currently calculated using these numbers (bad)
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/caa94993-32d0-4d6f-8e32-497fd06517ec)
Should be calculated using these numbers
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/51c3d75b-f7d2-4aa5-93f6-d0662afcdb36)

Venting losses Unit Conversions
LP vent & Condensate Flash vent don't seem to have the correct unit conversions.  
For AT6
Low Pressure Vented steam loss should be 15.57 mmbtu/hr (not correct),  16251 MJ/hr (close, but not exact)
Condensate Flas loss should be  0.537 mmbtu/hr (not correct), 593 MJ/hr (not correct - weird huh)

In MJ/hr
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/ad64269a-8591-4474-b706-ac9566ad7ea5)

In MMBTU/hr
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/b59775c1-c4c0-402b-b2d2-80902f536ca4)

koay9f:
Also, it would probably help making the sankeys if we add a row:
"Returned Steam and Condensate"
 {deaeratorFeedWater.massFlow * deaeratorFeedWater.enthaply* whatever unitconversions}  - loss.makeupWaterEnergy    (I think that's going to be the easiest way to do that...)

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/3e37b5f8-eb11-4d99-82e6-4d0243ab0e85)

koay9f:
If needed:
[https://app.zenhub.com/files/80439269/6fe54748-25f3-47b5-beee-6463e2dd5b9a/download](https://app.zenhub.com/files/80439269/6fe54748-25f3-47b5-beee-6463e2dd5b9a/download)

koay9f:
- [ ] Blowdown Losses using blowdown - post Heat Exchanger
- [ ] LP vent
- [ ] Condensate Flash
- [ ] Returned steam & condensate

Blowdown still seems to be incorrect

PR #4067, opened 2020-10-21
issue 4064

rmroot:
connects #4064 

fix vented steam losses and add returned steam and condensate energy calc
fix flash tank visual bug


rmroot:
also quick fix to import service..

nbintertech (approved):

Issue #4132, opened 2020-11-24
Validation updates to Charge Materials


nbintertech:
Changing some warnings to validation in new standalone calculator and assessment.

koay9f:
Where did this one come from?
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/105d19ba-0537-46e6-9f5b-99a811ad1ca7)

Are there any others that got added that I should double check my dumb-ness?

Issue #4118, opened 2020-11-18
Charge Materials


koay9f:
Same as the others

koay9f:
Description:
Preheating of load or charge material by using heat of flue gases is one of the most efficient method of heat recovery and energy efficiency improvement.  The load or charge material is heated in the front section of a furnace or oven where hot gases transfer heat to the load before leaving the furnace.  In some cases a separate preheater is installed in which the load is heated by circulating hot flue gases before they are discharged. Load preheating also offers an opportunity to increase production or throughput if required.  Actual energy savings or fuel input reduction depends on combustion efficiency or available heat of the heating system.

koay9f:
Short description:
Estimates savings from preheating or changing how the load is heated.

koay9f:
Move Available Heat to after fuel costs

Add a line to separate this from the charge material specific fields.  

Add ability to have more than one  charge material.  Op hours, fuel cost, available heat do not need repeating, just everything between (including) Select Type and Additional Heat Required
If it has the ability to add multiples in the Assessment, it needs it here.

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/36b296ae-3d90-4bee-a86e-808d6914bae1)

PR #4149, opened 2020-12-08
Issue 4118

nbintertech:
connects #4118 #4132 

- Add charge material standalone calc
- Convert assessment warnings to validation where possible
- Use standalone calc form services in assessment


src/app/calculator/furnaces/charge-material/charge-material.component.html
```diff
@@ -0,0 +1,223 @@
...
+			  <h3>Baseline</h3>
+			</div>
+			<div class="d-flex flex-column mx-2 mb-4">
+			  <!-- app form -->
+			  <app-energy-form [selected]="baselineSelected" [isBaseline]="true" [settings]="settings"></app-energy-form>
+				<div class="form-group">
+				  <label class="small" for="materialType">Select Type</label>
+				  <select [disabled]="!baselineSelected" class="form-control" id="materialType" (change)="changeMaterialType()" [(ngModel)]="materialType"
+					 (focus)="focusField('chargeMaterialType')">
+					<option>Solid</option>
```
rmroot, line 42: I think these <option> need to have a [ngValue] property set

rmroot (changes requested):
Need to add scroll (height) to help/results panel, scrolling gets messed up when that panel has overflow. Melting point with savings suggestions open is good for testing

Should probably always have the "Select Type" dropdown disabled in modification since we want baseline and modification to be the same

rmroot (approved):

Issue #4113, opened 2020-11-13
WasteWater: Update names and help


rmroot:
use documents from Susana to update names and help text

PR #4163, opened 2020-12-15
Issue 4113

rmroot:
connects #4113 

updated field names, validation and minor bug fix


nbintertech (changes requested):
The BiomassYeild references are consistently spelled, though looks like a typo - should be Y**ie**ld?

rmroot:
> The BiomassYeild references are consistently spelled, though looks like a typo - should be Y**ie**ld?

Variables misspelled but that's how it is in the suite as well. Just going to leave it, I verified everything forward facing is correctly spelt.

Issue #3754, opened 2020-05-27
Fan Traverse Tweaks 2


koay9f:
May need to split this into multiple issues, but putting it here for now
More help text is needed - Ron will help (EDIT - about what?)

The results table should be in order of flow (left to right on the diagram)
EDIT: so the table should be 3a, 3b, 3c, 4, a, 2, 5 (5 might be before two if conditions are right)

~The planes don't have the right numbers on them (Ron will help and get back to us)~
EDIT: The planes numbers in ISO and AMCA are not the same. Add help text (see below)

EDIT: Ignore below this, moving TO DOs to a new comment

S-tube pitot tube types require a static pressure correction
* let users enter corrected static pressure (no change, but add some help texts)
* somewhere on the page, have a table the same size as the velocity pressure table where they can enter all the static pressures.  Final static pressure = average of all the entries)

redesign plane 3 pages.  the order is wonky (EDIT - how?)

"Velocity in the cross section would be helpful" - I think add a left hand side result with the velocity, if we can calc w/o calc-ing everything
In plane 3 - add help text for AMCA recommendations for number of traverse readings (EDIT - see below for functions in the suite)

koay9f:
putting table order in a separate issue.

koay9f:
Help Text Changes:

Plane 1 - is fine

Plane 2 -  ```Plane 2 is the fan outlet flange. If the evase was included in the Rated Fan Performance, Plane 2 can be located either upstream or downstream of Plane 5, however downstream is uncommon.```

Plane 3 - ```Plane 3 is the plane used for the Pitot-static traverse for purpose of determining flow rate. This plane is "Plane 5" in the ISO 5802 standard. [Picture]```

Plane 4 - ````Plane 4 is a plane upstream of fan used to establish fan inlet Plane Static Pressure. This plane is "Plane 3" in the ISO 5802 standard.````

Plane 5 - ```Plane 5 is a plane near the outlet used to establish fan outlet Plane 2 Static Pressure. If the evase was included in the Rated Fan Performance, this might be upstream or downstream of Plane 2.  This plane is "Plane 4" in the ISO 5802 standard. ```

koay9f:
Motor Input Power (input and result) should be in kW

Change name of tab "Fan Shaft Power" to "Motor Power" (maybe)

koay9f:
Total Issue To Do 
- [ ] Reorder fields
[...]

PR #4177, opened 2020-12-28
Issue 3754

nbintertech:
connects #3754 

Remaining issues from 3754
- Add "Calculate Static Pressure" functionality
- Motor Input Power (input change to kW (fix any unit conversions, etc.)
- Add "Fan Shaft Power" Result on page (bold left side result fine)
- Use settingsLabel
 
Please check the removed 53-66 pressure-readings-form.ts - tested in FSAT and standalone I couldn't find a case where this would be true.

### DIFF
--- a/src/app/calculator/fans/fan-analysis/fan-analysis-form/plane-data-form/pressure-readings-form/pressure-readings-form.component.ts
+++ b/src/app/calculator/fans/fan-analysis/fan-analysis-form/plane-data-form/pressure-readings-form/pressure-readings-form.component.ts
@@ -37,7 +37,8 @@ export class PressureReadingsFormComponent implements OnInit {
         this.setPlaneData();
         this.updateData();
       }
-    })
+    });
+    this.save();
   }
 
   ngOnDestroy() {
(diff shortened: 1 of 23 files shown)


src/app/calculator/fans/fan-analysis/fan-analysis-form/plane-data-form/pressure-readings-form/pressure-readings-form.component.ts
```diff
@@ -93,8 +88,31 @@ export class PressureReadingsFormComponent implements OnInit {
     this.save();
   }
 
+  setStaticPressure() {
+    let row;
```
rmroot, line 92: add types and a comment or two about what's being done here. (I should have added earlier -sorry)

rmroot (changes requested):
invalid-style

Styling on this is messed up, it's on ng-invalid class don't know why it's on the label.. I also think we should have the mdash red if its calculated

rmroot (approved):

Issue #4183, opened 2021-01-04
SVI calculator


koay9f:

koay9f:
[https://app.zenhub.com/files/80439269/820116bd-df35-458d-9c4e-c5447a8d7285/download](https://app.zenhub.com/files/80439269/820116bd-df35-458d-9c4e-c5447a8d7285/download)

nbintertech:
Calculator finished for the moment... waiting on feedback from original calculator creator/expert.

PR #4235, opened 2021-01-21
Issue 4183 b

nbintertech:
connects #4183 

- adds waste water SVI calculator


nbintertech:
Not including until after 1/25 release

Issue #4236, opened 2021-01-21
PH standalone calculator visual bug


nbintertech:
Divider/container background is visible for calculators with multiple losses
![divider.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/130cc2bf-0a7b-4001-9c69-5463446776ba)

koay9f:
Still there for
opening (2 losses)
Atmosphere (3 losses)
Wall (2 - 3 losses)
Charge Mat (2 losses)

nbintertech:
Below is the change we're including for the current release. We'll need to look at changing the layout to fix it completely - Should we create a new issue or push this back to inbox?
![divider bug.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/f8a778b2-c650-4e24-935a-35840afbb224)

koay9f:
I'm going to create a new issue but put it in the icebox.  It's not too bad I think and it sounds like it is not worth the time investment right now.

PR #4244, opened 2021-01-25
Issue 4236

nbintertech:
connects #4236 

fit-content height when multiple losses.


nbintertech:
Closing for better solution

Issue #3839, opened 2020-07-08
Leakage Standalone


koay9f:

koay9f:
Air leak losses occur when colder air leaks into a furnace or oven because the inside pressure in the furnace is less than the outside pressure.  This air is ultimately discharged as part of hot exhaust gas or flue gas. Energy required to heat the air leaked into the furnace/oven has to be supplied by the burner.  The air leaks can be reduced or eliminated by adjusting the furnace inside pressure (less negative or ideally zero) and/or reducing or eliminating the opening size.  An automatic pressure control system is often used for large installations and manual pressure adjustment is used for smaller installations.  The most cost effective method is to eliminate or reduce opening size if possible. This calculator can be used to calculate energy savings and cost savings by taking any one of the actions mentioned above.   Actual energy savings or fuel input reduction depends on combustion efficiency or available heat of the heating system.

koay9f:
Short description:
Estimates losses due to gases leaking out of or into the furnace

koay9f:
looking good except: 
I think the Fuel Use calculation is a problem for other calculators too. (not the gross loss, that seems to be working)
This seems to be related to multiple losses. I'm guessing it is not using the right operating hours for the second loss?
Is a problem for: Wall, Opening, Gas Leak, Atmo
Charge Materials is okay and it is not applicable to Flue Gas

koay9f:
still has a problem with operating hours & multiple losses

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/5392961a-fbc5-4434-9b5c-9b61fc51cd0b)

PR #4272, opened 2021-02-01
Issue 3839

nbintertech:
connects #3839


src/app/calculator/furnaces/leakage/leakage.service.ts
```diff
@@ -0,0 +1,291 @@
...
+    if (currentModificationData) {
+      currentModificationData.splice(i, 1);
+      this.modificationData.next(currentModificationData);
+    }
+  }
+  
+  addLoss(treasureHours: number, modificationExists: boolean) {
+    let currentBaselineData: Array<LeakageLoss> = JSON.parse(JSON.stringify(this.baselineData.getValue()));
+    let index = currentBaselineData.length;
+    let hoursPerYear = treasureHours? treasureHours : currentBaselineData[index - 1].hoursPerYear;
```
rmroot, line 197: index-1 breaks when no baseline data. `index-1` always big red flag

rmroot: Suggestion.. 

instead of continually accessing the object in the array using `baselineData[index-1]`

add a third optional parameter to `initDefaultLoss` that is a LeakageLoss.
pass `baselineData[0]` 
if it exists, use it to set the `fuelCost` and `availableHeat` of the new loss

rmroot (changes requested):

rmroot (approved):

Issue #4302, opened 2021-02-22
Add update-units-modal and warning to system basics in fsat, psat, ssmt


nbintertech:
When units of measure are changed in assessment system-basics tab we should see a confirmation popup (update-units-modal) and a warning message. this will replace the existing warning below.

This already exists for PHAST and should implemented the same way in FSAT, PSAT, and SSMT.
 
In project see files:
- phast > system-basics component
- update units modal

Existing warning message
![replace message.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/7ab170d7-8e65-4728-bc84-6b8406490973)

New message and Modal below
![new message.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/72cd1173-d5ed-4696-a69d-2dd9ce6a2c3d)

![changeunits.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/e9c61af4-ced3-4cc2-bfc3-9180b4593552)

![update-units modal.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/4486504e-927c-420b-92e6-c435d4e4b63a)

nbintertech:
Need to add modal to treasure hunt as well

koay9f:
The yellow/green message box works, but the modal doesn't pop up.  (It also no longer pops up for Process Heating, which it did when I tested the PH one earlier today...)

Also, I think unit conversion is not currently implemented for treasure hunt, so it probably needs a special modal for now (something like "At this time MEASUR cannot convert the data fields for an in progress Treasure Hunt"

PR #4335, opened 2021-03-08
Issue 4302

rahuldino:
Connects #4302 
Implemented update units modal for PSAT, SSMT and FSAT.

### DIFF
--- a/src/app/fsat/fsat.component.ts
+++ b/src/app/fsat/fsat.component.ts
@@ -17,6 +17,7 @@ import { FanMotorService } from './fan-motor/fan-motor.service';
 import { FanFieldDataService } from './fan-field-data/fan-field-data.service';
 import { FanSetupService } from './fan-setup/fan-setup.service';
 import { SettingsService } from '../settings/settings.service';
+import { ConvertFsatService } from './convert-fsat.service';
 
 @Component({
   selector: 'app-fsat',
@@ -68,6 +69,7 @@ export class FsatComponent implements OnInit {
   showAdd: boolean;
   isModalOpen: boolean;
   openModSub: Subscription;
+  modalOpenSubscription: Subscription;
   calcTab: string;
   calcTabSubscription: Subscription;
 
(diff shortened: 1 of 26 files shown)


src/app/fsat/fsat.component.ts
```diff
@@ -154,11 +157,6 @@ export class FsatComponent implements OnInit {
         this.modificationIndex = undefined;
       }
     });
-
-    this.modalOpenSubscription = this.fsatService.modalOpen.subscribe(isOpen => {
```
nbintertech, line 158: We need to keep this sub. It's published to by other modals in child forms (see fan-field-data.ts).

src/app/fsat/fsat.component.ts
```diff
@@ -427,4 +424,22 @@ export class FsatComponent implements OnInit {
...
+    this.showUpdateUnitsModal = true;
+    })
+  }
+
+  closeUpdateUnitsModal() {
+    this.showUpdateUnitsModal = false;
+  }
+
+  selectUpdateAction(shouldUpdateData: boolean) {
+    if(shouldUpdateData == false) {
```
nbintertech, line 440: If shouldUpdateData == true, we need to convert the input data. See same method in phast.component.ts. If possible, pull out the conversion logic from system-basics --> updateData() and move it to a service method like is done in ConvertPhastService -->convertExistingData(). Otherwise pull the logic out and place in fsat.component.

nbintertech:
All comments on FSAT apply to PSAT and SSMT also. In the system-basics html for each section we also need to replace the original message. See PHAST --> system-basics lines 6-16

nbintertech (changes requested):
All comments on FSAT apply to PSAT and SSMT also. In the system-basics html for each section we also need to replace the original message. See PHAST --> system-basics lines 6-16.

Happy to get on a call if we need to clarify anything

nbintertech:
Closing to put up new branch/PR

Issue #4331, opened 2021-03-05
Fan Report changes


nbintertech:
parent #4328

Input Summary - Motor Power (and probably Motor Current) has a "input" for the mods, should just be "- - - " for mods. Probably also true for pumps.
Input Summary - if use "known efficiency" do "- - - " for fan type
~Input Summary - always says fan type says "radial tip"~

koay9f:
If use "Known Fan Efficiency" Not sure if that is enough of a trigger, it might have to be "did not click optimize"
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/7d514366-e3f5-42a8-8212-7bff104a5c31)

This is what "I Know the Fan Eff" or "not optimizing" looks like
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/d39a259c-40da-403a-88eb-868864aaf045)
This is what "optimizing" looks like
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/379f4731-f26f-4d98-8e40-b98305483d2d)

koay9f:
For all modifications, "Load Estimation Method" and "Motor Power" (or "Motor Current") should be " - - " 
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/e97c438b-eed6-4098-8af7-ef945eb689cb)

koay9f:
Ignore the third point

PR #4343, opened 2021-03-10
Issue-4331

rahuldino:
connects #4331 

made changes to the fan report input summary Fan section for fan type and in the Field Data section for load estimated method and motor power.


src/app/fsat/fsat-report/input-summary/field-data-summary/field-data-summary.component.html
```diff
@@ -96,7 +96,7 @@ <h5>Field Data
               </td>
               <td *ngFor="let mod of fieldData.modifications" [ngClass]="{'indicate-report-field-different': checkDiff(fieldData.baseline.loadEstimatedMethod, mod.loadEstimatedMethod, 'loadEstimatedMethodDiff', 0)}">
                 <span *ngIf="mod.loadEstimatedMethod == 0">
```
rmroot, line 98: Should be &mdash; for either loadEstimatedMethod == 0 or loadEstimatedMethod == 1 so you can remove the `<span *ngIf`  and just put &mdash; &mdash;

rmroot (commented):
Need to apply these updates to psat report as well, in that case we are looking at pump_style == 11

rmroot (approved):

PR #4369, opened 2021-03-15
Issue 4312

rmroot:
connects #4312 

connects https://github.com/ORNL-AMO/AMO-Tools-Suite/issues/546

adds new fields for modification energy calculations


rmroot:
Should be noted need a local suite for testing until new version of suite is release

src/app/calculator/compressed-air/compressed-air-pressure-reduction/compressed-air-pressure-reduction.service.ts
```diff
@@ -135,11 +146,22 @@ export class CompressedAirPressureReductionService {
   generateExample(settings: Settings, isBaseline: boolean): CompressedAirPressureReductionData {
     let proposedPressure: number = 90;
     let pressure: number = 100;
+
+    //TODO: DEFAULTS
```
nbintertech, line 150: There are a couple TODOs here and in mockTreasurehunt

nbintertech (changes requested):

src/app/calculator/compressed-air/compressed-air-pressure-reduction/compressed-air-pressure-reduction.service.ts
rmroot: Good catch!

nbintertech (approved):

PR #4396, opened 2021-03-22
Issue 4302

nbintertech:
connects #4302 

- Adding update units modal to all assessments


rmroot (changes requested):
Need treasure hunt implemented.

Something strange going on with conversion for PSAT. Using example Fluid Temp goes from 68 degrees F to -21.48 degrees C. 

PHAST if you change units and convert not using the modal and then change units again the message doesn't show up again.

Similar for FSAT. Basically just double check change, convert, change, convert works as expected

rmroot (approved):

Issue #4361, opened 2021-03-11
Move FSAT Specific Heat Ratio  


nbintertech:
Parent #4328

Move Field "Specific Heat Ratio" from `Field Data` to `Fluid`

PR #4397, opened 2021-03-22
Issue 4361

nbintertech:
connects #4361 

Move specific heat ratio to fan fluid form

### DIFF
--- a/src/app/fsat/fsat-fluid/fsat-fluid.component.html
+++ b/src/app/fsat/fsat-fluid/fsat-fluid.component.html
@@ -31,6 +31,21 @@
       </span>
     </div>
 
+    <div class="form-group"
+      [ngClass]="{'indicate-different': isSpecificHeatRatioDifferent(), 'invalid':gasDensityForm.controls.specificHeatRatio.invalid}">
+      <label for="{{'specificHeatRatio_'+idString}}">Specific Heat Ratio (&gamma;)</label>
+      <input name="specificHeatRatio" type="number" min="0" step="any" class="form-control"
+        formControlName="specificHeatRatio" id="{{'specificHeatRatio_'+idString}}" (input)="save()"
+        (focus)="focusField('specificHeatRatio')" onfocus="this.select();" [readonly]="!selected">
+      <span class="alert-danger pull-right small"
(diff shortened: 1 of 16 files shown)


src/app/fsat/fsat-fluid/fsat-fluid.component.html
```diff
@@ -31,6 +31,21 @@ <h3>Fluid</h3>
...
+    <div class="form-group"
+      [ngClass]="{'indicate-different': isSpecificHeatRatioDifferent(), 'invalid':gasDensityForm.controls.specificHeatRatio.invalid}">
+      <label for="{{'specificHeatRatio_'+idString}}">Specific Heat Ratio (&gamma;)</label>
+      <input name="specificHeatRatio" type="number" min="0" step="any" class="form-control"
+        formControlName="specificHeatRatio" id="{{'specificHeatRatio_'+idString}}" (input)="save()"
+        (focus)="focusField('specificHeatRatio')" onfocus="this.select();" [readonly]="!selected">
+      <span class="alert-danger pull-right small"
+        *ngIf="gasDensityForm.controls.specificHeatRatio.invalid && !gasDensityForm.controls.specificHeatRatio.pristine">
+        <span *ngIf="gasDensityForm.controls.specificHeatRatio.errors.required">Value Required</span>
+        <span *ngIf="gasDensityForm.controls.specificHeatRatio.errors.max">Value can't be less than
```
rmroot, line 43: "Value can't be less than max" think it should say "Value can't be greater than"

rmroot (changes requested):
I think this will break old assessments. Need to map the old object to the new spot

src/app/shared/helper-services/update-data.service.ts
```diff
@@ -56,6 +56,10 @@ export class UpdateDataService {
             assessment.fsat.fieldData.inletVelocityPressure = 0;
             assessment.fsat.fieldData.usingStaticPressure = true;
         }
+
+        if(assessment.fsat.fieldData['specificHeatRatio']) {
```
rmroot, line 60: Lets add  `&& !assessment.fsat.baseGasDensity.specificHeatRatio`
 to the check..

Also should be updating modifications?

rmroot (changes requested):

rmroot (approved):

PR #4465, opened 2021-04-26
Issue 4357: Add Dashboard and Home buttons to settings page

JustinCIS:


rmroot (changes requested):
Everything is working but aesthetically it needs some work...
Move the buttons into the card and use similar styling as we have on the dashboard

image

rmroot (approved):
Looks good!

Issue #4488, opened 2021-05-06
WebAssembly: Remove Process Heat Addon


rmroot:

rmroot:
TODO: List Bindings for testing

nbintertech:
phastAddon:
- [ ] fixtureLosses
- [ ] gasCoolingLosses
- [ ] gasLoadChargeMaterial
- [ ] liquidCoolingLosses
- [ ] liquidLoadChargeMaterial
- [ ] viewFactorCalculation
- [ ] openingLossesQuad
- [ ] openingLossesCircular
- [ ] solidLoadChargeMaterial
- [ ] wallLosses

- [ ] leakageLosses
- [ ] flueGasLossesByVolume
- [ ] flueGasLossesByMass
- [ ] flueGasCalculateExcessAir
- [ ] flueGasCalculateO2
- [ ] flueGasByMassCalculateExcessAir
- [ ] flueGasByMassCalculateO2
- [ ] atmosphere
- [ ] slagOtherMaterialLosses
- [ ] auxiliaryPowerLoss

- [ ] energyInputEAF
- [ ] exhaustGasEAF
- [ ] energyInputExhaustGasLosses
- [ ] efficiencyImprovement
- [ ] energyEquivalencyElectric
- [ ] energyEquivalencyFuel
- [ ] flowCalculations
- [ ] o2Enrichment
- [ ] flueGasByVolumeCalculateHeatingValue
- [ ] flueGasByMassCalculateHeatingValue

processHeatAddon:
- [ ] waterHeatingUsingSteam
- [ ] airHeatingUsingExhaust
- [ ] cascadeHeatHighToLow
- [ ] waterHeatingUsingExhaust

PR #4522, opened 2021-05-21
Issue 4488

nbintertech:
connects #4488 

Add PHAST API service


nbintertech:
Just noticing Omer's PR for #602 that changes heatCascade endpoints. Will update here when that gets merged

Issue #4166, opened 2020-12-16
Audit Fan/Pump help text


nbintertech:
Remove any min/max tables that don't provide useful information (i.e. below). Expected min/max don't appear to be working. Double check anything else suspect.

![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/8aa5cf40-179c-403f-83ca-88f6b097fe70)

![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/23fe0a0d-6142-4ac0-b0bd-925163710520)

nbintertech:
The two above are examples, but I believe these min/max tables occur all over in the app. Since it's a larger issue feel free to work on this in pieces if needed.

koay9f:
It's not really a table, but it still should be removed:
Pumps.  Full Load Amps (It is likely also in Fans, and anywhere else FLA is a field)
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/621378d9-ca83-4fae-b6b9-ba191d685bc2)

This one can be removed also
Pumps Electricity Cost (may be in other places also)
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/6dbc800f-f515-41ab-97ce-a7200d372546)

PR #4555, opened 2021-06-08
removed min/max tables that don't provide useful information

shivasaurus:
Connects #4166


nbintertech:
See issue again - I've updated information.

shivasaurus:
All remaining useless min/max tables should be removed now

nbintertech (approved):

Issue #2851, opened 2019-03-05
Steam Losses table Zero values


koay9f:
When the loss = 0 the line doesn't show up on the table.

This is good for when the loss *shouldn't be there* (a turbine loss for a turbine that isn't present), but not good if there is a 100% eff turbine or a turbine with zero flow.  may want to rethink how it works

koay9f:
[https://app.zenhub.com/files/80439269/24ebc6d9-8828-4851-a942-88a429c19737/download](https://app.zenhub.com/files/80439269/24ebc6d9-8828-4851-a942-88a429c19737/download)

I think this is only a concern for turbines.  Can you send a list (in slack, I won't notice it here) of any of the lines in that table that can "disappear" and what condition makes them appear/disappear?

An example is in the ppt above. Or I can send you an assessment via slack if you want.  just let me know

koay9f:
I think this is causing the Sankey to crash.  May need to pull in Nick.

Below is an example.  you will need to change from .txt to .json

[https://app.zenhub.com/files/80439269/0ca983c4-2271-4d52-beaa-336774aff918/download](https://app.zenhub.com/files/80439269/0ca983c4-2271-4d52-beaa-336774aff918/download)

PR #4568, opened 2021-06-11
turbines now display even if loss is 0

shivasaurus:
Connects #2851


src/app/ssmt/calculate-losses.service.ts
```diff
@@ -234,10 +234,10 @@ export class CalculateLossesService {
       highPressureHeader: 0,
       mediumPressureHeader: 0,
       lowPressureHeader: 0,
-      condensingTurbineEfficiencyLoss: 0,
-      highToMediumTurbineEfficiencyLoss: 0,
-      highToLowTurbineEfficiencyLoss: 0,
-      mediumToLowTurbineEfficiencyLoss: 0,
+      condensingTurbineEfficiencyLoss: -1,
```
rmroot, line 237: I like the idea here but it's gonna cause some issues in other areas of calculations if we set the values to -1 like calculate totalTurbineLosses(); 

I suggest, modifying the `SSMTLosses` object by adding four booleans that you use for show/hiding turbine. Use those booleans in the `*ngIf` statement.

They will end up matching the `.useTurbine` booleans of the `turbineInput` object in the `calculateLosses()` function.

I'm happy to hop on a call to explain further.

src/app/ssmt/ssmt-report/losses-summary/losses-summary.component.ts
```diff
@@ -79,25 +79,25 @@ export class LossesSummaryComponent implements OnInit {
   }
 
   checkCondensingTurbine(loss: SSMTLosses) {
-    if (loss.condensingTurbineEfficiencyLoss || loss.condensingTurbineUsefulEnergy) {
+    if (loss.condensingTurbineEfficiencyLoss >= 0|| loss.condensingTurbineUsefulEnergy) {
```
rmroot, line 82: Here you would check the new booleans added

rmroot (changes requested):

src/app/ssmt/ssmt-report/losses-summary/losses-summary.component.ts
shivasaurus: These changes have now been pushed

rmroot (approved):

Issue #2383, opened 2018-09-21
PSAT / FSAT reports notes


koay9f:
Where should the notes be shown in the report?
Result Data? at the bottom like PHAST?

koay9f:
Add notes to the report like in PHAST

nbintertech:
![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/60aa8bcc-7437-404b-a810-eecc854d3d21)

![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/6c5a73c0-612d-488a-8cff-e807dcba59da)

![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/b4c4bf0d-52d2-4e82-a25d-508e2dedf32a)

![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/92e173a1-0c78-40ec-af42-2156e3b9c275)

PR #4570, opened 2021-06-14
Issue 2383: PSAT & FSAT modification notes on report and rollup

rhernandez-intertech:
connects #2383

### DIFF
--- a/src/app/fsat/fsat-report/results-summary/results-summary.component.html
+++ b/src/app/fsat/fsat-report/results-summary/results-summary.component.html
@@ -231,42 +231,19 @@
       </tr>
     </tbody>
   </table>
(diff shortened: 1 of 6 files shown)


nbintertech:
I think we can improve on the output-summary html/ts (and the same in psat). Take a look at phast-report/executive-summary.component lines 156-168. That's all we'll need for markup. To do that, the ts needs to be changed to handle most of the logic to build our modification notes array. Start by searching the project for the 'buildSummaryNote' and 'buildNoteObject' methods. We'll borrow these methods to do the same in fsat/psat. Give these references a look and lets follow up via chat/call

nbintertech (changes requested):

nbintertech (approved):

Issue #4569, opened 2021-06-13
TH Integration - Vent Steam to heat water


nbintertech:
Vent steam -
This one is weird. It will have fuel (possibly two different ones) & water
Still no BL vs. mod

Results:
BL energy = energy savings from heater + energy savings from boiler
mod energy = 0
savings = energy savings from heater + energy savings from boiler

BL water = water savings
mod water = 0
savings = water savings

for costs: don't bother with water treatment costs, just water costs
** for energy: there will be a fuel for boiler and a fuel for HX, options are NG and "other fuel"

koay9f:
Boiler Utility Type should be the second field
Both `Cost of Fuel` fields should be filled by the default for the utility type selection

Same issue air heating
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/d9e0b289-73c5-4236-8cbc-1ac3327910ae)
Just used "generate example" but for other fuel, switched to other fuel.

PR #4575, opened 2021-06-16
Issue 4569

nbintertech:
connects #4569 

- Add vent steam to heat water to TH
- set heat cascade inputs undefined on destroy when in th


rmroot (changes requested):
Looks good but the title header should go the full width of the screen to be consistent with the other calculators. (same for air heating)

rmroot (approved):

Issue #4527, opened 2021-05-24
Yet another CA Pressure Reduction Tweak


koay9f:
Review Request:
Make so user can specific if using Rated values or measured values.
Need a toggle.
Will hide some fields, change field names and small algorithm difference that can be handled on the desktop side.
will be adding more details soon.

koay9f:
[Uploading... pressure reduction.pptx]

koay9f:
UPDATE: math is different, what is in ppt will not work:

Have a toggle for Measured/Rated

For MEASURED:
set BL & mod "Rated Pressure" = BL "Operating Pressure"
Hide "Rated Pressure"
Let's leave the field name "Operating Pressure" alone, it is good.

For BOTH:
Move Compressor Power to Bottom.  Hide in mod.  (and maybe always hide Rated Pressure in the mod...)

koay9f:
I think something extra got hidden

Rated - Baseline should have 3 pressures: Rated, Operating & Atmospheric

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/db2539d6-68b1-4946-a0e6-627a3d145f9b)

PR #4593, opened 2021-06-21
4527

nbintertech:
connects #4527 

- Adding power type for measured or rated
- Hide fields depending on power type


rmroot (changes requested):
Do we need to handle old treasure hunt compressed air reductions by setting some default?

rmroot (approved):

Issue #4355, opened 2021-03-11
Fan Setup Rearrangement


koay9f:
Combine `Fluid` and `Fan` Tabs to more closely mirror Pumps?

PR #4598, opened 2021-06-23
combined fan and fluid tabs

shivasaurus:
Connects #4355


nbintertech:
Everything looks good, with a few modifications:

- Let's display the fan section on top of fluid and emulate the help text/header from pump and fluid page.
- Add the same subheader pattern to pump and fluid page (everything after drive field belongs to fluid).

nbintertech (changes requested):
Everything looks good, with a few modifications:

- Let's display the fan section on top of fluid and emulate the help text/header from pump and fluid page.
- Add the same subheader pattern to pump and fluid page (everything after drive field belongs to fluid).

nbintertech (approved):

Issue #4591, opened 2021-06-21
Process Cooling Psychrometric descript


koay9f:
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/3603f186-d107-4efd-946b-4ee26aebdc7e)

“Calculate the properties of moist air based on formulas published in ASHRAE Fundamental Handbook.”

remove word "base" where possible

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/ed7efb78-c14a-44a0-ba0c-bd4408bb98f6)

koay9f:
Part two is good.  Part 1 is not done:
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/1619499a-7ea2-4be8-a31f-85298127c9e6)
Text with the red line should be
`Calculate the properties of moist air based on formulas published in ASHRAE Fundamental Handbook.`

Issue #4586, opened 2021-06-18
Help text titles in stand alones


koay9f:
Make sure help text titles match the calc title in stand alones
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/c778c978-720b-42fd-b8b3-b6c222f8d21c)

Issue #4581, opened 2021-06-17
Pump calc help


koay9f:
Pump Achievable Efficiency & Specific speed

copy help text from pump assessment for pump type

For Achievable Eff - help has both field at once instead of field specific

koay9f:
Specific Speed is perfect
Pump Achievable Efficiency - no longer has both help texts, but it also doesn't have any.
So need the help text for pump type and flow rate

Issue #4558, opened 2021-06-09
Restore Settings Language


koay9f:
Restore System Settings
Restoring system settings will provide options for resetting all application settings to their default values, returning all example assessments to their original state, deleting all user-created assessments/pre-assessments, and deleting all user-created materials. This cannot be undone!

See if we can come up with some better language. Make it clear that this is not an all or nothing thing and hitting the big red button won't kill all your assessments

PR #4604, opened 2021-06-30
cmerch branch

c-merchant:
Connects #4586 
Connects #4591 
Connects #4581 
Connects #4558


nbintertech (changes requested):
There are a number of files where plotly chart fonts are changed to roboto. I didn't see that change in any of the issues listed on PR. There's also a change that looks like it got snuck in from operating costs issue. 

Let me know what's going on with these and we can work to resolve.

c-merchant:
Connects #216

@nbintertech The roboto changes were part of my progress on Ticket #216, but the operating costs change did just sneak in there from the other ticket I'm working on. I plan to make branches for each issue in the future after closing out these tickets to avoid the intersection of such things.

nbintertech (approved):

Issue #4602, opened 2021-06-28
Build config to move wasm files automatically


nbintertech:
Setup config to move client.js and client.wasm from assets into dist.

PR #4613, opened 2021-07-06
Run as webapp fixes (WASM 4602 4614)

nbintertech:
connects #4602 #4614 

- Output client/amo-tools-suite to dist directly
- Add window ref to polyfills.js
- Check if running electron app before running update check

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -51,7 +51,7 @@
               {
                 "glob": "**/*",
                 "input": "./node_modules/amo-tools-suite/bin/",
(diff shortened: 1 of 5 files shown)


nbintertech:
- Updated commit fixes not being able to build web and desktop version from the same config. 
- Double checked need for polyfill change. https://github.com/angular/angular-cli/issues/9827#issuecomment-386154063

rmroot (approved):

Issue #4143, opened 2020-12-07
Improve Copy-table Strings


nbintertech:
We're concatenating copy table strings in Motor Drive, Lighting Replacement and a few other places when there are multiple copy tables. 

Fix all instances. This will probably just be adjusting HTML structure.  Search app for "this.tableString ="

nbintertech:
@rhernandez-intertech let's meet on this before you start. Need a better idea of what/where is working.

PR #4618, opened 2021-07-06
Issue 4143

rhernandez-intertech:
connects #4143

### DIFF
--- a/src/app/calculator/motors/motor-drive/motor-drive-table/motor-drive-table.component.ts
+++ b/src/app/calculator/motors/motor-drive/motor-drive-table/motor-drive-table.component.ts
@@ -11,14 +11,18 @@ export class MotorDriveTableComponent implements OnInit {
   results: MotorDriveOutputs;
 
   @ViewChild('copyTable', { static: false }) copyTable: ElementRef;
-  tableString: any;
+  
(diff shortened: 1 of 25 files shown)


src/app/calculator/motors/motor-drive/motor-drive-table/motor-drive-table.component.ts
```diff
@@ -11,14 +11,18 @@ export class MotorDriveTableComponent implements OnInit {
   results: MotorDriveOutputs;
 
   @ViewChild('copyTable', { static: false }) copyTable: ElementRef;
-  tableString: any;
+  
   @ViewChild('copyTable2', { static: false }) copyTable2: ElementRef;
+
+  tableString: any;
```
nbintertech, line 17: This existed before your changes, but here and any other files you touched where tableString (or any other property) use any - change type to string or whatever the type should be.

nbintertech (changes requested):
If you search the project for '.hidden-copy-table' you'll see two positioning styles used (absolute/fixed). We want the one with fixed. Please change this for the sat properties table you made and fix any of the others. This allows them to be hidden without being rendered over other elements. You may have to play with the positioning, notice the top value for some.

nbintertech (approved):

Issue #4514, opened 2021-05-13
Import Excel sheet/CSV into editable Traverse Holes/Insertion points


nbintertech:
Ideally the user can navigate up, down, left, right, return etc on all traverse hole inputs, similar to an excel sheet. There is already a library being used to import csv in data-exploration tool. See if this will work.

![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/6ac4ff45-cfd0-40d2-ad2c-31be2ae15f4f)

nbintertech:
Not  high priority at the moment, but a longer project to keep a steady pace on. Start poking around at this issue,

PR #4622, opened 2021-07-08
Added Excel import for Traverse Holes

shivasaurus:
Connects #4514


src/app/calculator/fans/fan-analysis/fan-analysis-form/plane-data-form/pressure-readings-form/pressure-readings-form.component.ts
```diff
@@ -48,6 +56,70 @@ export class PressureReadingsFormComponent implements OnInit {
...
+    }
+  }
+
+  importFiles() {
+    this.fileReferences.forEach(fileReference => {
+      let excelTest = /.xlsx$/;
+      if (excelTest.test(fileReference.name)) {
+        const reader: FileReader = new FileReader();
+        reader.onload = (e: any) => {
+          /* read workbook */
```
nbintertech, line 82: - Use let instead of var
- Remove commented ifelse block
- Comment (or rename) what we're matching in ln 100 regex

nbintertech (changes requested):
Nice work here. If you haven't already, also send Kristina a screen of the page. We may want more visual division from the rest of the form. You'll want to match the font size as well

rmroot (approved):

nbintertech (approved):

Issue #4460, opened 2021-04-23
Compare two existing pump/fan


koay9f:
In expert view
A question/radial button:
Compare two existing pumps/fans or what-if scenario?
O Two existing pumps/fans (have field data for both)
O What-if scenario

If What-if scenario - algorithm as is
If Two existing - use the baseline algorithm twice

koay9f:
When create first scenario, no way to specify that it is a "compare two pumps" scenario:
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/b609e37d-2732-4408-957f-d77e4d6c3e08)

Showing energy/cost savings even if no changes have been made
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/4dd63475-7807-48ae-92d2-8805c19c52e7)
Might just be better in this case to not show any savings between the two pumps... I'll get back to you.

koay9f:
Visually, looking good.
Baseline results seems to be copying Scenario 1 instead of using the baseline though
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/104f91d1-9207-45de-9526-fa9d92440349)

I checked, it is not doing this in whatif

Also something weird going on with the saving...
looks like it might just be motor power, Load Estimation Method and Motor Current
(I'm guessing it is b/c before, those fields weren't in the assessment, so it just isn't saving them yet)
![issue4460.gif](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/174fefec-76e1-4426-b1fb-ca8ed521096b)


Weirdly, fan's looks like it is doing okay

nbintertech:
@rhernandez-intertech pull from branch `issue-4460-c` to look at this. I took care of the basellne copying scenario 1 issue. Please investigate the saving/badge issue.

PR #4629, opened 2021-07-13
Issue 4460: Fans and Pumps compare two as baseline or what-if scenario

rhernandez-intertech:
connects #4460

### DIFF
--- a/src/app/psat/modify-conditions/modify-conditions.component.html
+++ b/src/app/psat/modify-conditions/modify-conditions.component.html
@@ -35,6 +35,7 @@
           [(psat)]="psat.modifications[modificationIndex].psat" (saved)="save()" [settings]="settings"
           [modificationIndex]="modificationIndex"></app-pump-fluid>
         <app-motor *ngIf="modifyTab == 'motor'" [selected]="modifiedSelected"
+        [baseline]="psat.modifications[modificationIndex].psat.inputs.whatIfScenario == false"
           [(psat)]="psat.modifications[modificationIndex].psat" (saved)="save()" [settings]="settings"
           [modificationIndex]="modificationIndex"></app-motor>
         <app-field-data *ngIf="modifyTab == 'field-data'" [(psat)]="psat.modifications[modificationIndex].psat"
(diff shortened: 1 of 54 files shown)


nbintertech (changes requested):
You've got a solid start here. Things to keep in mind:
- Look over var/method names and make sure they are meaningfully named to represent what we're doing
- Look at update-data.service.ts and consider the properties you're adding might affect old assessments and test with one.
- Once the user has selected a 'what-if' compare method, where are we setting the algorithm to calculate the second equipment as baseline?

src/app/shared/helper-services/update-data.service.ts
```diff
@@ -63,11 +63,19 @@ export class UpdateDataService {
         if(assessment.fsat.modifications){
```
nbintertech, line 63: Don't forget to add a similar check for psat

src/app/shared/models/psat.ts
```diff
@@ -13,6 +13,7 @@ export interface PSAT {
   operatingHours?: OperatingHours;
   valid?: PsatValid;
   existingDataUnits?: string;
+  whatIfScenario?: boolean;
```
nbintertech, line 16: Put the whatIfScenario property on PsatInputs instead of PSAT, so you don't need to pass it in to psat-report -> getResults(). Same with field-data-service.ts or anywhere else where you've added it as an argument and we already have PsatInputs. In Field-data.ts you won't need to pass in whatIfScenario to get the form... this.baseline will be accurate since it will be based off the whatIfScenario passed from the parent.

src/app/psat/pump-fluid/pump-fluid.component.html
```diff
@@ -42,12 +42,12 @@ <h3>Pump</h3>
       </span>
     </div>
 
-    <div class="text-center small help-text" *ngIf="psatForm.controls.pumpType.value == 11 && !baseline">
+    <div class="text-center small help-text" *ngIf="psatForm.controls.pumpType.value == 11 && (!baseline && !compareTwoPumpsAsBaseline)">
```
nbintertech, line 45: We shouldn't need to use the compareTwo.. anywhere

src/app/psat/modify-conditions/modify-conditions.component.html
```diff
@@ -19,9 +19,37 @@ <h3>Baseline</h3>
       </div>
     </div>
 
+    <!-- Modifications Start / Compare Two Pumps as baseline -->
```
nbintertech, line 22: This may be what's causing some of the form display issues. You have two modification blocks where `psat.modifications[modificationIndex].psat.whatIfScenario == false` is being checked. You only need one modification block, and that property doesn't need to be checked. See my example branch

nbintertech (approved):

PR #4635, opened 2021-07-14
Issue 4632

c-merchant:
All entries with IDControlType = 11 have been removed


rmroot:
Need the PR to be into compressed-air branch and not develop

Issue #4633, opened 2021-07-14
CompressedAir: Field by Field help


rmroot:
most help text is in "Names in MEASUR" excel file from Alex

koay9f:
Need some more fields.  Alex will provide ASAP.

koay9f:
Tab level help text:
System Info - `Enter data related to entire compressed air system`
Inventory - `Enter data relating to all individual compressors within the system`
Day Types - is fine

Assessment - 
Use the list on the left to discover potential opportunities for energy and cost savings within your system.
---------------
MEASUR will automatically reallocate your systems flow to match your compressor ordering established in the baseline, either via automatic cascading (according to the full load pressure of each compressor) or sequenced within the system profile. 

Need EEM help text. Each EEM has EEM level help text like in Process Heating
Reduce Air Leaks - Models the impacts of reducing air leaks
Leak Flow - Total estimated leak load of the system, in acfm [N/m3 in metric]. You can use the "Air Leak Survey" in the calculators tab to get a better estimate. 
Leak Reduction - Percentage of the leak flow that will be fixed

Improve End Use Eff - Models air flow reductions for specific hours, achieved by reducing the amount of air required by end users.
Air Flow Reduction type - Can choose either fixed or variable reduction. Fixed means the same reduction will be applied to all hours, variable allows the user to set the reduction for each hour.
Airflow reduction - Amount of airflow that will be reduced
Airflow reduction (for variable) - Amount of airflow that will be reduced during that hour

Reduce System Air pressure - Models the impact of reducing system pressure.
Avg System Pressure Reduction - Pressure reduction for the system. This affects your performance points. 

Use Unloading Controls - Models the impact of changing your compressor control scheme or type, allowing your compressor to unload or shutdown to save energy at low demand.
Select Compressor - select the compressor who's unloading controls you which to change.  Changing this field does not erase the data you entered for another compressor. The asterisk indicates a compressor that's control type was changed. Restore Baseline resets the data for the compress you have selected back to the baseline data.
Unload Point, %C - use BL help text
Automatic shutdown timer - use BL help text
Proposed Control Type - use BL help text + Control Types options are based on your Baseline control type. The asterisk indicates the Baseline control type
Performance points - use BL help text

Use Automatic Sequencer - Models your system to add a sequencer or change your compressor sequencing
Target Pressure - use BL help text
Variance - use BL help text

[...]

PR #4673, opened 2021-07-21
Issue 4633

rhernandez-intertech:
connects #4633


nbintertech:
Created new branch with fixed git history

PR #4718, opened 2021-08-02
Add validation icon for CA collapsed inventory

nbintertech:


src/app/compressed-air-assessment/inventory/control-data/control-data.component.html
```diff
@@ -1,6 +1,9 @@
 <label class="group-label" (click)="toggleCollapse()">
     <span class="fa fa-chevron-right" [ngClass]="{'fa-chevron-right': contentCollapsed, 'fa-chevron-down': !contentCollapsed}"></span>
     Controls
+    <span class="pull-right">
```
rmroot, line 4: should put the *ngIf on the outside span so that there isn't an empty span (apply to all)

rmroot (changes requested):

rmroot (approved):

Issue #3554, opened 2020-02-10
Allow users to set $ or $k


koay9f:
Make costs a unit (will help for future internationalization as well)
Can change between $ and $k
Global setting, Assessment Type Setting & Individual assessment setting

koay9f:
Pumps: - need to do unit conversion approval - that's odd
Field Data
Explore Ops - missing
Modify All - missing
Results = good
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/35158b5a-1382-4af5-a27b-14bac38b1c90)

fans:
Field data - good
Explore Ops - good
Modify All - good
Results - good

Pumps:
Explore Ops
Modify All
Results

Pumps:
Explore Ops
Modify All
Results

Pumps:
Explore Ops
Modify All
Results

shivasaurus:
Didn't complete any additional work on this.

nbintertech:
Can close #2803 if all good.

koay9f:
Still not doing a conversion in ...
Pumps: System Setup >Field data > Annual Cost
Pumps: Report > Result Data > Annual & Savings - I think is doing a double conversion (if you look at the example annual cost should be 152 $k, but is instead showing as "00")
Treasure Chest -
mostly good. "Other Cost Savings" % is still wonky (same as above pic) and report is a bit wonky
$
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/58d21bf8-ad0f-4d05-8048-2364077c7e71)
$k
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/216b491b-877f-46ce-a1e2-c47b0692d8c3)
Also in Opportunity Summary, Payback Details, Report graphs (but since hte probably will probably be fixed before those are generated, I imagine that it will just filter though...)



Folder cards - showing results as "Annual Energy Costs: $152" and "Max Costs Savings $19" (instead of $152k and $19k). Adding wrong in the summary cards also.
Roll up - the k is in the units, but it still assumes that everything in the folder is in the same currency units.  need to do something similar to if we have mixed metric & imperial units


Other weird...
Pumps: Assessment>Explore Opps>Annual Cost displays with a 0 in the "tens" slot (so 6 shows as 06)

nbintertech:
Turning off this currency setting until we can prioritize time to make the required changes to the TH section

nbintertech:
[...]

PR #4729, opened 2021-08-04
Currency change on assessments (switches between $ and $k)

shivasaurus:
Connects #3554


rmroot (changes requested):
Any existing settings object from before this update is going to have "$ - US Dollar" as it's value,

indexedDb/settings-db.service.ts has a function called "checkSettings()" that is called whenever a settings object is retrieved. Should add logic in there that updates "$ - US Dollar" to "$"

rmroot (approved):

Issue #1873, opened 2018-06-11
Additional graph to add for electrotech


koay9f:
Energy input: Electrical vs. Chemical/Fuel

koay9f:
![image](https://user-images.githubusercontent.com/25389736/41236607-8baebeaa-6d5f-11e8-8e82-608e16167b2a.png)

PR #4739, opened 2021-08-09
added additional graph

shivasaurus:
Connects #1873

### DIFF
--- a/src/app/phast/phast-report/report-graphs/report-graphs.component.ts
+++ b/src/app/phast/phast-report/report-graphs/report-graphs.component.ts
@@ -22,13 +22,16 @@ export class ReportGraphsComponent implements OnInit {
   @Input()
   printGraphs: boolean;
 
+  hasDeliverInput: boolean = true;
+
   allChartData: Array<{
     name: string,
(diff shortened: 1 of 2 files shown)


src/app/phast/phast-report/report-graphs/report-graphs.component.ts
```diff
@@ -76,16 +84,33 @@ export class ReportGraphsComponent implements OnInit {
...
       barChartValues: valuesAndLabels.map(valueItem => {return valueItem.value}),
       modification: modification,
-      valid: isValid
+      valid: isValid,
+      deliverValuesLabels: deliverValuesLabels
     })
   }
 
+  getDeliverValuesAndLabels(phast: PHAST): Array<{ value: number, label: string }> {
+    let results: PhastResults = this.phastResultsService.getResults(phast, this.settings);
```
rmroot, line 100: in here we are calling `getResults()`, it also calls that function in the `getValuesAndLabels()` function. both called by `addChartData` 

call `getResults` in the `addChartData` function and pass the results into the other two functions please

rmroot (changes requested):

rmroot (approved):

Issue #4537, opened 2021-06-07
Calculators with multiple baseline BL/Mod setups UI bug


nbintertech:
Both baseline and mod columns should be the same height. 

This shows up in calculators with multiple baseline and mods (ex. process heating/furnace calcs). We use the .scroll-item class to set scroll for this window and programmatically set window height onInit(). This may need to change.

![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/bcfdd869-ef9e-4f1a-bf2e-f52a437f6f41)

nbintertech:
Still some remaining issues in SSMT and calculators

PR #4747, opened 2021-08-10
Calcs in Process Heating with multi Bl's and mod's UI fixed

rhernandez-intertech:
connects #4537


nbintertech:
Looks good. There are a few other calculators with bl/mod setup that need this, if you could please seek them out and update.

nbintertech (approved):

Issue #4746, opened 2021-08-10
Cleanup Perf Points Decimals


nbintertech:
This is for default compressor data showing in performance point forms. May require some finagling due to the same instance of data being used to perform calculations

From Alex:
>  kW = #.# Flow(acfm) = ##  Pressure (psig) = #.#

PR #4757, opened 2021-08-12
Issue 4746

nbintertech:
connects #4746 

Cleanup perf point decimals

### DIFF
--- a/src/app/shared/models/compressed-air-assessment.ts
+++ b/src/app/shared/models/compressed-air-assessment.ts
@@ -129,6 +129,7 @@ export interface CompressorInventoryItem {
     name: string,
     description: string,
     isValid?: boolean,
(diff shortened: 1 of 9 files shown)


src/app/shared/models/compressed-air-assessment.ts
```diff
@@ -129,6 +129,7 @@ export interface CompressorInventoryItem {
     name: string,
     description: string,
     isValid?: boolean,
+    isUpdatingFromGeneric?: boolean,
```
rmroot, line 132: I don't think this is used

rmroot (approved):

src/app/shared/models/compressed-air-assessment.ts
nbintertech: Removed leftovers

Issue #4681, opened 2021-07-22
Bugs in the psychrometric calc


koay9f:
In Fan module and stand alone calc...

From Wei
Bugs:
1.	On the right, the absolute pressure is probably incorrect. It should be the sum of Barometric Pressure and Static Pressure. {from koa - actually, I think just the label is incorrect and maybe the unit conversions.
In Imperial - absolute pressure's unit label should be "in Hg"
In metric - I'm guessing there is something wrong with the unit conversions... I think it is trying to do in H2O to Pa when it should be doing in Hg to Pa or something else weird...)
2.	On the right, the Barometric Pressure is probably incorrect. It should be 101,325. I think the unit was treated as KPa. (from koa - change unit conversion to kPa instead of Pa)

Ideas for Improvement:
1.	All pressure units, it might be good to make them consistent. Per Ron, he prefers Pa. (if it is kPa --> Pa, if it is kPaa --> Paa)
2.	Only Static Pressure on the left is gauge pressure, so we do not have to add “a” after “Pa”, but all other pressure values should be absolute or “Paa”. Thank you! {from koa - change Barometric pressure and Absolute Pressure on the right to Paa)

 ![fanscreenshot.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/fc3d3119-8eb2-4e6c-8b82-09e66be76ebd)

koay9f:
Unit conversion in the form and results are not correct.
Imperial w/ correct Pa conversions
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/6973dbcf-9b9e-43c5-9849-8c774bc51b30)
Barometric 29 in Hg = 98205.3 Pa
Static: 13 in H2O = 3386.39 Pa (correct)
Abs Pressure result: 30 in Hg = 101592 Pa

When you switch to metric, Static pressure converts fine, but Barometric's unit conversion is converting to kPaa and I don't know what is going on with Abs Pressure
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/09e7d59d-11fd-4ae0-80cd-fbaea5d09bb5)

And if you try to match the label for barometric, you get yelled at b/c it thinks you are putting a number in kPa
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/7a721156-843e-4c00-9178-1830061a7dc3)

[...]

PR #4781, opened 2021-08-17
updated calculations for Bar P, Static P, and Abs P

DePauwJosh:
connects #4681


nbintertech (changes requested):
Are changes needed in the standalone calc (fan-psychrometric-form) also?

nbintertech (approved):

Issue #4736, opened 2021-08-06
Add Clear Bins button


nbintertech:
In weather binning calculator user should be able to keep current weather data/file in memory but clear all bins. 
![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/7c17123a-3752-48f9-a0c1-d2bbf61693aa)

PR #4795, opened 2021-08-24
Issue 4736

DePauwJosh:
connects #4736


src/app/calculator/utilities/weather-bins/weather-bins-form/bins-form/bins-form.component.ts
```diff
@@ -68,3 +75,7 @@ export class BinsFormComponent implements OnInit {
     }
   }
 }
+function setAutoBinCases() {
```
nbintertech, line 78: Don't think this is needed. Left from an autofill?

nbintertech (changes requested):

nbintertech (approved):

Issue #1428, opened 2018-03-16
Print Settings


koay9f:
Add print settings in general settings

Aeledbetter:
What kind of print settings?

koay9f:
[https://app.zenhub.com/files/80439269/8500a0c3-a6a3-4708-8317-8c721825c708/download](https://app.zenhub.com/files/80439269/8500a0c3-a6a3-4708-8317-8c721825c708/download)

Basically, the options for the print rollup are settings we can set in the global settings, but they don't have application in all modules (no "losses summary" for treasure hunt). But something like "Result Details" being checked off means that, by default, any module + roll up where you can print "Result Details" has that checked off

The ppt also has a few other updates we could do with printing.  let me know if any of them are too out of bounds and I can make another issue

PR #4796, opened 2021-08-24
Issue 1428: Print Settings

rhernandez-intertech:
connect #1428

### DIFF
--- a/src/app/settings/print-settings/print-settings.component.ts
+++ b/src/app/settings/print-settings/print-settings.component.ts
@@ -57,7 +57,7 @@ export class PrintSettingsComponent implements OnInit {
       this.settings.printWasteWaterRollup &&
       this.settings.printDetailedResults &&
       this.settings.printReportDiagram &&
-      this.settings.selectAll
+      this.settings.printAll
     );
   }
 
@@ -80,7 +80,7 @@ export class PrintSettingsComponent implements OnInit {
     this.settings.printWasteWaterRollup = this.selectAllOptions;
     this.settings.printDetailedResults = this.selectAllOptions;
     this.settings.printReportDiagram = this.selectAllOptions;
-    this.settings.selectAll = this.selectAllOptions;
+    this.settings.printAll = this.selectAllOptions;
     this.savePrintSettings.emit(true);
   }
 
--- a/src/app/settings/print-settings/print-settings.component.ts
+++ b/src/app/settings/print-settings/print-settings.component.ts
@@ -1,6 +1,7 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
 import { Subscription } from 'rxjs';
 import { PrintOptions } from '../../shared/models/printing';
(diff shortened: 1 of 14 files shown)


src/app/settings/print-settings/print-settings.component.ts
```diff
@@ -0,0 +1,189 @@
...
+  selectAllOptions: boolean = true;
+
+  printOptions: PrintOptions;
+  printOptionsSub: Subscription;
+
+  constructor(private printOptionsMenuService: PrintOptionsMenuService) { }
+
+  ngOnInit() {
+    this.checkSelectAll();
+    // this.printOptionsSub = this.printOptionsMenuService.printOptions.subscribe(val => {
```
nbintertech, line 27: Remove commented out code

src/app/settings/print-settings/print-settings.component.ts
```diff
@@ -0,0 +1,189 @@
...
+  settings: Settings;
+  @Output('savePrintSettings')
+  savePrintSettings = new EventEmitter<boolean>();
+
+  selectAllOptions: boolean = true;
+
+  printOptions: PrintOptions;
+  printOptionsSub: Subscription;
+
+  constructor(private printOptionsMenuService: PrintOptionsMenuService) { }
```
nbintertech, line 23: Don't need to inject PrintOptionsMenuService. This is initializing it at a global/shared scope and so it won't properly initialize and get global settings when print-options-menu.comp opens in fsat/psat etc..

src/app/settings/settings.module.ts
```diff
@@ -40,7 +43,8 @@ import { RouterModule } from '@angular/router';
         AnimatedCheckmarkModule
     ],
     providers: [
-        SettingsService
+        SettingsService,
+        PrintOptionsMenuService
```
nbintertech, line 47: Remove PrintOptionsMenuService and the import. Already part of another provider array. See other comment on print-settings.comp.ts

src/app/shared/print-options-menu/print-options-menu.service.ts
```diff
@@ -1,16 +1,22 @@
...
 
+  settings: Settings;
   printOptions: BehaviorSubject<PrintOptions>;
   showPrintView: BehaviorSubject<boolean>;
   printContext: BehaviorSubject<string>;
   showPrintMenu: BehaviorSubject<boolean>;
-  constructor() {
-    let initPrintOptions: PrintOptions = this.setAll(true);
+  constructor(private settingsDbService: SettingsDbService) {
+    if (!this.settings) {
```
nbintertech, line 16: this.settings isn't used anywhere and the check isn't needed since we're always grabbing the default print options from global settings

nbintertech (changes requested):
This doesn't currently work as far as I understand the issue. No global settings affecting report/rollup settings. Making the suggested changes should get it most of the way there, but more testing should be done.

2 things to check - I wasn't sure if intentional:
- Some of the bottom border lines are thicker on the global print settings
- Energy summary and losses summary can only be selected/unselected together

PR #4816, opened 2021-09-02
Compressed Air help text for explore ops

rhernandez-intertech:
connects #4633

### DIFF
--- a/src/app/compressed-air-assessment/results-panel/help-panel/help-panel.component.ts
+++ b/src/app/compressed-air-assessment/results-panel/help-panel/help-panel.component.ts
@@ -11,13 +11,21 @@ export class HelpPanelComponent implements OnInit {
 
   selectedTab: string;
   selectedTabSub: Subscription;
+  assessmentTab: string;
+  assessmentTabSub: Subscription;
+
+  mainTab: string;
   constructor(private compressedAirService: CompressedAirAssessmentService) { }
 
   ngOnInit(): void {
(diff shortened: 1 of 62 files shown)


src/app/compressed-air-assessment/results-panel/help-panel/help-panel.component.ts
```diff
@@ -11,14 +11,22 @@ export class HelpPanelComponent implements OnInit {
...
   ngOnInit(): void {
-    let mainTab: string = this.compressedAirService.mainTab.getValue();
-    if (mainTab == 'system-setup') {
+    this.mainTab = this.compressedAirService.mainTab.getValue();
+    if (this.mainTab == 'system-setup') {
       this.selectedTabSub = this.compressedAirService.setupTab.subscribe(val => {
         this.selectedTab = val;
       });
+    } else if (this.mainTab == 'assessment') {
+      this.assessmentTabSub = this.compressedAirService.assessmentTab.subscribe(val => {
```
nbintertech, line 27: Need to also check for and unsubscribe from assessmentTabSub in the ngOnDestroy method

nbintertech (changes requested):
Inside of a given Setup Prof section, all help fields are showing at the same time. If that was what was talked about - all good.

Take 20-30 mins to see if you can figure out what's causing this horizontal scroll. Not sure it was introduced by this PR but would be nice to fix. I've also seen it appear on the Setup Prof side
horiz scroll

Issue #4800, opened 2021-08-26
Add Bin table/matrix to Weather Binning Calculator


nbintertech:
Pull latest from cooling-and-weather. Add table where each bin is row and columns are upperbound, lower bound, hours

![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/82a748a3-bef7-47fa-8aa3-2f2eaf3be201)

koay9f:
can't see.  Looks like the table is under the graph
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/b996e41c-138e-4c7d-b449-c60344ac5d50)

PR #4823, opened 2021-09-07
Weather Bins result table

rhernandez-intertech:
connects #4800

### DIFF
--- a/src/app/calculator/utilities/weather-bins/weather-bins-table/weather-bins-table.component.ts
+++ b/src/app/calculator/utilities/weather-bins/weather-bins-table/weather-bins-table.component.ts
@@ -16,12 +16,7 @@ export class WeatherBinsTableComponent implements OnInit {
 
   constructor() { }
 
-  ngOnInit(): void {
-    
-  }
-
-  ngOnDestroy() {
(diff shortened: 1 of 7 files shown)


src/app/calculator/utilities/weather-bins/weather-bins-table/weather-bins-table.component.ts
```diff
@@ -0,0 +1,31 @@
...
+  styleUrls: ['./weather-bins-table.component.css']
+})
+export class WeatherBinsTableComponent implements OnInit {
+  @ViewChild('copyTable', { static: false }) copyTable: ElementRef;
+  tableString: any;
+
+  @Input()
+  inputData: WeatherBinsInput;
+
+  constructor() { }
```
nbintertech, line 17: Can remove these three methods not being used

nbintertech (changes requested):

nbintertech:
The table isn't styled _exactly_ like our other result tables, though I think the flat style looks good here. Look at other example tables or demo it like this for at the weekly.

nbintertech (approved):

Issue #4834, opened 2021-09-09
Export Data Explorer files


koay9f:
similar to the "load last instance" link. be able to export file and import

nbintertech:
Will split this issue up when I can get a better idea of scope

rmroot:
UI Implementation for export:  Add button to top right on banner using export icon from font awesome. This button should trigger a behavior subject (put in log-tools.services.ts) that is subscribed to in the base log-tool.component.ts. When true open a modal in that component.

![image.png](https://images.zenhubusercontent.com/5c9d19352250f9495b573154/ec9e7c0d-95c1-4633-918d-617de96d77ee)

- [ ] Add button icon (think about disabling when no data is loaded)
- [ ] Add behavior subject
- [ ] Add modal to base component and open/close it with behavior subject

Once these steps are complete we can implement the logic for exporting the data.

rmroot:
Import issue #4868

koay9f:
when click button it alternates between two things happening:
1. nothing
2. export data modal comes up, but cannot export as .json
![explorerbug.gif](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/cd14e77d-e0d8-42cc-ad04-9930d78c6585)

PR #4879, opened 2021-09-20
Issue 4834: Export Data Explorer files

rhernandez-intertech:
connects #4834

### DIFF
--- a/src/app/log-tool/export-modal/export-modal.component.html
+++ b/src/app/log-tool/export-modal/export-modal.component.html
@@ -9,25 +9,18 @@
                 </button>
             </div>
             <div class="modal-body pb-0">
-                <div class="export-content justify-content-center">
-                    <div class="pt-2" *ngIf="dataExists">
-                        Download Data from Last Use
-                        ({{data.modifiedDate | date:'long'}})
-                    </div>
-
+                <p class="small" *ngIf="dataExists">
+                    Download data from last use ({{data.modifiedDate | date}}) to be imported at a later time.
+                </p>
+                <div class="form-group">
+                    <label for="exportName"> Name your file:</label>
+                    <input type="text" class="form-control w-100" id="exportName" [(ngModel)]="exportName">
                 </div>
             </div>
             <div class="modal-footer justify-content-between">
                 <button class="btn btn-secondary" (click)="closeExportData()">Cancel</button>
-                <div class="d-flex">
-                    <div class="form-group d-flex mr-2">
-                        <label for="exportName"> Name your file:</label>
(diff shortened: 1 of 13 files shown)


src/app/log-tool/export-modal/export-modal.component.ts
```diff
@@ -0,0 +1,81 @@
...
+    });
+  }
+  ngOnDestroy() {
+    if (this.dataAvailableSub) {
+      this.dataAvailableSub.unsubscribe();
+    }
+  }
+
+  getExportData(){
+    this.dataAvailableSub = this.logToolDbService.previousDataAvailable.subscribe(val => {
```
rmroot, line 49: Looks like your using dataAvailable to show last modified date. The object LogToolDb has a propery "modifiedDate" you can use that on the "data" object and remove this subscription logic

src/app/log-tool/export-modal/export-modal.component.html
```diff
@@ -0,0 +1,40 @@
+<div bsModal #exportModal="bs-modal" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="exportModalLabel"
+    aria-hidden="true" [config]="{backdrop: 'static'}">
+    <div class="modal-dialog modal-lg export-modal">
```
rmroot, line 3: We don't need such a big modal. the "modal-lg" class can probably be like "modal-md"

src/app/log-tool/export-modal/export-modal.component.html
```diff
@@ -0,0 +1,40 @@
...
+        <div class="modal-content">
+            <div class="modal-header">
+                <h3>Export Data</h3>
+                <button class="close" aria-label="Close" (click)="closeExportData()">
+                    <span aria-hidden="true">&times;</span>
+                </button>
+            </div>
+            <div class="modal-body pb-0">
```
rmroot, line 11: I don't think we need the buttons here if we also have the buttons in the footer. I'd say just use the button in the footer and put a help text type of message in the body

src/app/log-tool/log-tool-banner/log-tool-banner.component.css
```diff
@@ -24,4 +24,13 @@
     width: 45px;
     background-color: white;
     border-radius: 5px;
+}
+
+button{
+    border: none;
+    background-color: none;
```
rmroot, line 31: this background style isn't being applied. you should use classes instead of html tags in .css, you may need to use "!important" for this, the style your overriding is in app.components.css which I think ends up taking priority without it

rmroot (changes requested):

rmroot (approved):

Issue #4868, opened 2021-09-17
Import Data Explorer files


rhernandez-intertech:
similar to the "load last instance" link. be able to import file

PR #4883, opened 2021-09-22
Issue 4868: Import .JSON files in Data Explorer 

rhernandez-intertech:
connects #4868


rmroot (changes requested):
Looking good. It's more related to the export issue but can be tackled here.. we should add some help text to the modal that's something like "Depending on the size of your data set, the exported file may be quite large and will take a few seconds." and then make sure that modal doesn't go away right away. The screen ends up just sitting there locked up while it builds the export file

rmroot (approved):

Issue #4877, opened 2021-09-20
Efficiency Improvement validation bug


nbintertech:
Probably a misspelling in template html
![eff improvement broke.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/f7777f8c-2b43-4efe-bcf5-2f699c18daa4)

PR #4901, opened 2021-09-27
Issue-4877: Efficiency Improvement validation bug

rhernandez-intertech:
connect #4877


rmroot (changes requested):
This doesn't work. Make sure you're testing your work before submitting a PR...

The problem is the less then validators are being set to null when the dependent value is undefined. Need to change the updateFormValidators function in the service to look like this...


image

rmroot:
> This doesn't work. Make sure you're testing your work before submitting a PR...
> 
> The problem is the less then validators are being set to null when the dependent value is undefined. Need to change the updateFormValidators function in the service to look like this...
> 
> image

Probably need to make that if a `value != undefined`, if it's 0 we still want it set

rmroot (approved):

Issue #4853, opened 2021-09-13
Charge Material and Flue gas 'Select Type' Dropdowns


nbintertech:
Dropdowns causing changed after checked error. Apply cd.detectChanges() like in other parts of the app or refactor.
![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/29c75b37-a03c-4901-a03b-fd4789ceaca6)

PR #4906, opened 2021-09-28
Issue 4853: Charge Material and Flue gas 'Select Type' Dropdowns bugs

rhernandez-intertech:
connects #4853

### DIFF
--- a/src/app/phast/losses/charge-material/charge-material.component.html
+++ b/src/app/phast/losses/charge-material/charge-material.component.html
@@ -64,8 +64,8 @@
       <div class="form-group">
         <label class="small" for="{{'type_'+lossIndex+idString}}">Select Type</label>
         <select class="form-control" id="{{'type_'+lossIndex+idString}}" [(ngModel)]="material.chargeMaterialType"
-          [disabled]="(!baselineSelected || lossesLocked) && !inSetup" (focus)="changeField('chargeMaterialType')"
-          >
+          (change)="saveLosses()" [disabled]="(!baselineSelected || lossesLocked) && !inSetup"
+          (focus)="changeField('chargeMaterialType')" (blur)="focusOut()">
           <option>Solid</option>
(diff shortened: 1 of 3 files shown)


rhernandez-intertech:
In Charge Material and Flue Gas in the PHAST an Expression-Changed-After-It-Has-Been-Checked error would occur when the material type changed from the "select type" drop down.
I am not sure what was causing the bug, other than something wasn't being rechecked after the material type changed. 
To fix the bug I implemented cd.detectChanges() in ngOnInit() and in changeField()

src/app/phast/losses/charge-material/charge-material.component.html
```diff
@@ -82,7 +82,7 @@ <h4>To add Charge Material you will need to delete your current modifications an
       [settings]="settings" (inputError)="setError($event)" [inSetup]="inSetup"></app-liquid-charge-material-form>
     <app-gas-charge-material-form *ngIf="material.chargeMaterialType == 'Gas'" (changeField)="changeField($event)"
       [(chargeMaterialForm)]="material.gasForm" (calculate)="calculate(material)" [baselineSelected]="baselineSelected"
-      [isBaseline]="isBaseline" (saveEmit)="saveLosses()" [lossIndex]="lossIndex" [settings]="settings"
+      [isBaseline]="isBaseline" (saveEmit)="saveLosses()" [lossIndex]="lossIndex" [settings]="settings" (inputError)="setError($event)"
```
rmroot, line 85: there is no inputError event emitter in the Gas Charge Material component

rmroot (changes requested):
It looks like adding the saveLosses() to the select dropdown is what fixed the issue. the added cd calls in this case isn't doing anything to solve the problem so it can be removed. I was unable to re-create the same bug in flue gas so it can also be removed there.

rmroot (approved):

Issue #4856, opened 2021-09-13
Explore Opps with Multiple Losses 


nbintertech:
![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/d6ee68d6-6d5f-44a7-929e-7b5ce51e9bf6)

rhernandez-intertech:
The error could not be recreated on the most recent version of Develop, however a new bug was found in Charge Martials.
 ![image (2).png](https://images.zenhubusercontent.com/60b7bed2e39f152beecc60b4/cd4bd340-e457-4b4f-bf52-075dd979c51a)
Charge showing an error with no validation

PR #4909, opened 2021-09-29
Issue 4856: Charge Martials form validation bug

rhernandez-intertech:
connects #4856


rhernandez-intertech:
The original error could not be recreated on the most recent version of Develop, however a new bug was found in Charge Martials. 
There was a validation bug in form caused by the Charge Inlet Temp and Charge Outlet Temp fields. When the same value was entered in both fields Charge Materials was showing an error but there was no validation from the form. To fix the bug I changed the validations of Charge Inlet Temp from the max being the value of Charge Outlet Temp to less than the value of Outlet Temp. I also added checkWarnings() to the Gas Material component and its service since it did not have one and Solid and Liquid did. I also modified the HTML files of the Charge Material Calculator since its services are shared with the PHAST losses and because the same bug was present there.

Issue #4888, opened 2021-09-23
Data Explorer: Last saved file bug


DePauwJosh:

koay9f:
what was the bug?  can't really test it....

DePauwJosh:
See issue-4818

PR #4911, opened 2021-09-30
Issue-4888

DePauwJosh:
connects #4888


nbintertech (changes requested):
Remove debuggers in both files please.

nbintertech (approved):

Issue #4744, opened 2021-08-10
Adjust Width of Results Panel


nbintertech:
Allow user to adjust width of results panel similar to how we can adjust left side navigation. This is for chiller staging (base branch `cooling-and-weather`. We may implement this app-wide eventually.

![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/7fca840d-6039-40b0-827c-f326959067a9)

koay9f:
I like this. please implement app wide.

PR #4920, opened 2021-10-01
Issue 4744: Adjust Width of Results Panel in Chiller Staging calc

rhernandez-intertech:
connects #4744


nbintertech (commented):
Looks good. If this is app-wide it will eventually need to be some sort of shared component that we pass in min/max x values and maybe manage the lifecycle of more intentionally. 

Not sure how far down the road that will be. Have Mark review and/or bring up in Thursday meeting.

rmroot:
Looks good to me. Will let Nick merge at his discretion.

Issue #4874, opened 2021-09-20
Air Leak Survey bugs


nbintertech:
1. Leak name does not reset after delete - Reproduce by deleting the name for a leak, click to edit another leak, then select the leak with deleted name, and add a new name.
2. Leaks table disappears - Delete all leaks, then add new. Might also be happening after some combination of example/reset

![name error.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/6413dc79-3536-45ef-baf3-5fc16ba3488b)

![table gone.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/5a7bb0c1-1937-43bc-b9b4-9a72438f3bbf)

PR #4922, opened 2021-10-04
bug that makes table disappear

DePauwJosh:
connects #4874

### DIFF
--- a/src/app/calculator/compressed-air/air-leak/air-leak.service.ts
+++ b/src/app/calculator/compressed-air/air-leak/air-leak.service.ts
@@ -80,9 +80,17 @@ export class AirLeakService {
     this.generateExample.next(false);
   }
 
-  deleteLeak(index: number) {
-    this.airLeakInput.value.compressedAirLeakSurveyInputVec.splice(index, 1);
-    this.airLeakOutput.value.leakResults.splice(index, 1);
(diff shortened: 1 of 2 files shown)


src/app/calculator/compressed-air/air-leak/air-leak.service.ts
```diff
@@ -80,9 +80,17 @@ export class AirLeakService {
     this.generateExample.next(false);
   }
 
-  deleteLeak(index: number) {
-    this.airLeakInput.value.compressedAirLeakSurveyInputVec.splice(index, 1);
-    this.airLeakOutput.value.leakResults.splice(index, 1);
+  deleteLeak(index: number, settings?: Settings) {
```
nbintertech, line 83: We'd normally disable the delete button if one item remains, but this calculator has an odd setup so lets stick with your solution.

- Method is only called from one place so settings doesn't need to be an optional param.
- If it's the last remaining leak, we're deleting the last leak from existing inputs and also reinitializing empty inputs. Should only need to reInit empty inputs. Take a look at what that method does if you haven't
- I'm not sure we need to resetData here. Changes should populate from initDefaultEmptyInputs.

nbintertech (changes requested):

nbintertech (approved):

Issue #5013, opened 2021-10-21
Copy to Folder 


rhernandez-intertech:
similar to bulk to move folder, but copy selected items to new location

PR #5018, opened 2021-10-25
Issue 5013: Copy to Folder

rhernandez-intertech:
connects #5013


rhernandez-intertech:
I added a call to init() method at the end of toggelIsAllSelected() method to update the directory, this was implemented in the bug fix commit

nbintertech:
I'm not sure if the new cancel button and text will meet basic accessibility guidelines. Change to white text otherwise using a transparent/white background like we do elsewhere is fine.

nbintertech (approved):

Issue #5008, opened 2021-10-20
Modal Size - results overflow


koay9f:
I've only found this here, but the results is overflowing the modal.
Process Heating>Fuel-fired>Flue Gas> Solid/Liquid Fuel > Calculate Moisture in combustion air modal

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/67f50c5c-3f64-4964-8ec2-7d2727d830a9)

rmroot:
Fixed modal layout. When verifying make sure to check the units for the inputs and results are correct, was previously based on fan settings

koay9f:
units look good. could probably generally clean it up (more like inputs for stand alone psychrometic than fan density,) but it gets the job done. would want to talk to Sachin first so will close this issue for now.

PR #5028, opened 2021-10-29
Issue-5008: Modal Size - results overflow

rhernandez-intertech:
connects #5008


rmroot (changes requested):
Should find a solution that doesn't require a hard coded height value.

nbintertech:
Still needs a dynamic height.  If we merged as a temporary fix we should update the issue to explain why it can't resize effectively or what needs to be done. That said, there are other modal examples to check out that are resizing correctly. It may come to refactoring the markup.

rmroot:
created new branch/PR. closing

Issue #4807, opened 2021-08-30
Placement of Cost & opHours Waste Water


nbintertech:
Broken out from 4349:

some assessments have the cost and opHours in an "Operations" page. This is good. make others like this.
For wastewater - the two "modeling conditions" - if that won't screw things up. (SRT might, so it can stay if it needs to)

PR #5100, opened 2021-11-09
Issue 4807: Waste Water Operations Page

rhernandez-intertech:
connects #4807


nbintertech (changes requested):
This needs further testing. The app is crashing when Aerator performance or mainTab of Assessment is clicked. This happens when there is a difference in suite input objects vs desktop side.

- Go over logic again for TimeIncrement. My assumption is it should be everywhere that MaxDays and the other properties are. It's undefined from the inputs you're sending to the suite - possibly causing crash.
- Some code blocks related to system-settings are commented out. If they are being removed, just remove them. If we ever need to go back in time, git has everything we need.

nbintertech (approved):

Issue #2869, opened 2019-03-07
SSMT Print Report Tweaks


koay9f:
Need to go over report and iron some things out
d
* Input Data - Turbines "Operation Value 1" and 2 can we get those to show what was used (mass flow or power generation)

* waterfall with mods looks weird.  maybe two waterfalls on one page per mod?

rmroot:
Operation Value 1/2: How do you want it labeled if baseline is Mass Flow and modification is Power Generation?

koay9f:
Bother.

A- If all would be same (both Flow, flow range or balance header OR both power, power range or balance header) --> Mass flow or Power Generation (as applicable)
B - come up with something better than `Operation Value`  maybe just Flow/Power

PR #5108, opened 2021-11-10
Issue-2869: SSMT Print Report Tweaks

rhernandez-intertech:
connects #2869


nbintertech (changes requested):
Do we still need the (unit) or (Other unit) on the row label if we have it displayed on the data? I don't know what was discussed.

nbintertech (approved):

Issue #4924, opened 2021-10-04
Add CO2 Emissions Results


nbintertech:

koay9f:
emissions are not being calculated for modifications
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/0767e858-2371-41e1-8098-b1bf8ebee769)

PR #5110, opened 2021-11-12
Issue 4924

nbintertech:
connects #4924 
Add assessmentCo2Savings shared comp, integrate with psat module and global settings

- Kept a bunch of comments for Josh and Rachel, incase using for reference.
- Made co2 settings part of the assessment/psat interface since it's really data being used for results - not settings;


rmroot (approved):
Looks good. Would just double check the shared components are not assessment specific like "emitUpdatePsatCo2SavingsData"

Issue #5038, opened 2021-11-01
Roll up report additional graphs


koay9f:
Something to make it actually "roll up" everything
* Total annual Electricity, Natural gas (be careful including the TH and steam in this)
* Graph with % energy use accounted for (aggregate pumps, fans, compressors, etc.- so like 50% of your energy is for process heating, 5% fans, 45% compressors)

koay9f:
for ref:
Pumps, Fans, compressed air, wastewater == "electricity"
Process Heating depends on what type for each assessment (fuel or steam = "fuel", electrotech = electricity)
Steam = fuel

rhernandez-intertech:
The graphs are now in uniformed units, the common report unit. and the graphs update when the common report unit is changed

PR #5120, opened 2021-11-18
Issue-5038: Roll up report additional graphs

rhernandez-intertech:
connects #5038


nbintertech (changes requested):
The scaling and layout here needs to be reworked. Specifically, remove extra whitespace, scale the images appropriately, and try to maintain original style of the rollup cards in the first row. If we need to restructure markup a little that's probably ok if Kristina is approving of the end result.
image

PR #5122, opened 2021-11-19
Issue 4183 

nbintertech:
connects #4183 

Talked to Kristina - has been done for awhile and we can merge this in for the time being (Tom approved).


rmroot (changes requested):
Double check the sidebar/calculator order. Changes are made in this PR and I know that whatever they were in was intentional and they keep getting screwed up. Might be all good, just take a second look to verify

nbintertech:
Updated. Used Calc list from the most recent verified PR for it. Weird - not sure if there was some implication those mins would change

rmroot (approved):

Issue #5093, opened 2021-11-08
Compare 2 pumps/fans part 2


koay9f:
* Hide explore opps. this is confusing, should just see "modify all conditions"
* put choice in report somewhere

PR #5124, opened 2021-11-19
Issue-5093: Compare 2 pumps/fans part 2

rhernandez-intertech:
connects #5093

### DIFF
--- a/src/app/fsat/explore-opportunities/explore-opportunities-form/explore-opportunities-form.component.html
+++ b/src/app/fsat/explore-opportunities/explore-opportunities-form/explore-opportunities-form.component.html
@@ -11,7 +11,8 @@
     <div class="checkbox">
       <label class="group-label" for="modifyCalculationMethod">
         <input type="checkbox" id="modifyCalculationMethod" (focus)="focusField('modifyCalculationMethod')"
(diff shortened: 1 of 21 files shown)


nbintertech (changes requested):
Errors below when creating new assessments in fans and pumps, filling out setup, clicking assessment.
pumps
new assessment

nbintertech (changes requested):
There are currently a few things going on here:
- Creating a new FSAT, clicking the Ex Opps button, then going to scenarios - the scenario radio buttons are disabled. 
- Creating a new PSAT, clicking Ex Opps button, then going to scenarios and selecting compare two, I'm returned to ex opps with some disabled opps but not all. If I go over to Modify Conditions, the Ex Opps tab is disabled.
- In either new assessment type if I click modify conditions first, I'm still getting the same error from the screens above

Please take a look at these and let's talk so I can be on the same page with what intended functionality is here

nbintertech (changes requested):
- Let's tidy up the message a bit to match others. Use the yellow warning messages for example.
- Both fsat/psat need further testing. Ex.. there's an error in fans when you select an explore opps opportunity and then change your setup to compare two. in fsat-results-panel.ts you'll need to add another condition in ngOnchanges() to check for fsat changes and call getResults(). That may take care of it
create-mod-edit-ex-opps

nbintertech (approved):

Issue #5118, opened 2021-11-18
Motor Inventory: Add copy Motor functionality 


AlexBotts:
Under catalog, add the same copy motor functionality that we have in compressed air. 
![image.png](https://images.zenhubusercontent.com/5e4ffc1aa541aecb7b4865d7/dfd44ae6-a664-4702-9e6e-90f7ee638bd5)
![image.png](https://images.zenhubusercontent.com/5e4ffc1aa541aecb7b4865d7/86123fd9-b967-4687-b783-2ffe590e32fb)

PR #5131, opened 2021-11-23
Issue-5118: Motor Inventory: Add copy Motor functionality

rhernandez-intertech:
connects #5118


nbintertech (changes requested):
There is a method 'openConfirmDeleteModal' being used. No modal is opening

nbintertech (approved):

Issue #4923, opened 2021-10-04
Add CO2 Emissions Results


nbintertech:
We need to add CO2 Emissions results to PHAST. Look at existing CO2 savings calculator and the pumps > operations implementation as guides. Unlike pumps, we don't want to use the shared component here since PH has other energy types. Otherwise the implementation is pretty similar.

Here's a high level breakdown of what it looks like to hook this up:
- Add CO2 savings fields to the PHAST operations page
- Call e-grid.service methods to init CSV data so the zipcode field can do lookup
- The CO2 savings fields should be populated from the PHAST CO2SavingsData OR if it doesn't exist, populate from global settings
- Find where the results are calculated and add a calculation method to calculate from the fsat.CO2SavingsData property
- Display CO2 emissions output and savings in results

PHAST energy types:
	If fuel-fired energy type or steam energy type:
			- No zipcode field
			- Fuel select dropdown that populates emission based on fuel

nbintertech:
Please aim to have this in PR by 12/3. If there are any hurdles reach out to Mark and I so we can help wrap up.

rhernandez-intertech:
also adding field by field help text

koay9f:
* Don't need the "savings" result in any of the "right side results" (system setup or assessment (explore or expert))
Also (just to mess with you more), this result is for "hourly" results, so don't need the `x operatingHours` part for this. (the annual numbers in the Exe Summary are good)
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/2c5d5c2e-a19a-4111-ae55-4699e710d134)
So the result here should be 2498 kg/hr (unless there is some kind of rounding going on on this page that I don't remember)

* Unit conversion
I'm guessing you are doing a conversion from MMBtu to GJ for the emission factor when you should be doing (1/MMBtu) to (1/GJ) {test, set conversion to 10 kg/MMBtu, should be 9.47 kg/GJ, not 11.1}. Would it help if we made a new unit in the converter that was for emissions factors instead of trying to just use energy?
Let me know and I can make an issue and help get it set up.

Everything else looks good. Tested Electricity and other than the general issue above with the results, they look good. Tested steam, it looks good, but I imagine it has the same unit conversion problem as fuel-based.

koay9f:
* need to get rid of the kg to tonne conversion
[...]

PR #5147, opened 2021-12-03
Issue 4923: Add CO2 Emissions Results to PHAST

rhernandez-intertech:
connects #4923


nbintertech (changes requested):
Missing a global settings implementation. Has a few errors and some styling cleanup is needed. Sending word doc with comments and screens.

src/app/phast/losses/operations/co2-savings-phast/co2-savings-phast.component.ts
```diff
@@ -0,0 +1,362 @@
...
+      }
+      this.form.patchValue({
+        totalEmissionOutputRate: selectedFuelOutputRate
+      });
+      this.form.controls.userEnteredModificationEmissions.patchValue(false);
+    }
+  }
+  
+  setSubRegionData() {
+    this.egridService.getAllSubRegions();
```
nbintertech, line 245: This should probably be called in phast.component (see the pumps implementation). CSV data won't be read in time here - this is why when you create an electrotech assessment with a zipcode it doesn't populate an emissions output rate

src/app/phast/losses/operations/co2-savings-phast/co2-savings-phast.component.ts
```diff
@@ -0,0 +1,362 @@
...
+      modificationCo2SavingsData.zipcode = this.co2SavingsData.zipcode;
+      // IMPORTANT - the below vales were missing so bl values were not copied over to mod
+      modificationCo2SavingsData.energyType = this.co2SavingsData.energyType;
+      modificationCo2SavingsData.fuelType = this.co2SavingsData.fuelType;
+      modificationCo2SavingsData.energySource = this.co2SavingsData.energySource;
+      this.phastCO2SavingsService.modificationCo2SavingsData.next(modificationCo2SavingsData);
+    }
+  }
+
+  convertCO2SavingsData(co2SavingsData: Co2SavingsData): Co2SavingsData{
```
nbintertech, line 336: I don't believe there's a scenario where the conversions in here will be executed. emission output rate is in the correct unit when the user enters it, and you're converting it when it comes from the csv also. I don't think you need this. Take a second look

src/app/phast/convert-phast.service.ts
```diff
@@ -19,6 +19,8 @@ import { Settings } from '../shared/models/settings';
 import { DesignedEnergy } from '../shared/models/phast/designedEnergy';
 import { MeteredEnergy } from '../shared/models/phast/meteredEnergy';
 import { OperatingCosts } from '../shared/models/operations';
+import { Co2SavingsData } from '../calculator/utilities/co2-savings/co2-savings.service';
+import { settings } from 'cluster';
```
nbintertech, line 23: Remove this import

nbintertech (changes requested):
Please also remove any comments and debuggers that aren't necessary. Test through all scenarios after changes are made. New assessment with electric/fuel/etc. All modifications are being updated etc...

nbintertech (approved):

Issue #5157, opened 2021-12-07
Add CO2 saving field for Electricity in TH  


rhernandez-intertech:
Add Total Emission Output Rate field to Energy Usage and Cost page in TH under Electricity 
the field will be populated based on emission rate calculated from zip code, the default value will come from CO2 global settings. 
under the field there will be a link to a modal where the user the can change the zip code being used to calculate the total emission output rate (the shared assessment CO2 savings component can be used for this modal) 
  ![image.png](https://images.zenhubusercontent.com/60b7bed2e39f152beecc60b4/eb01c352-beb2-46c8-ba11-251f456396ac)

PR #5167, opened 2021-12-10
Issue 5157: Add CO2 saving field for Electricity in TH

rhernandez-intertech:
connects #5157

### DIFF
--- a/src/app/treasure-hunt/operation-costs/operation-costs.component.html
+++ b/src/app/treasure-hunt/operation-costs/operation-costs.component.html
@@ -37,6 +37,15 @@
             <span class="input-group-addon units">$/yr</span>
           </div>
         </div>
+        <div class="form-group" *ngIf="treasureHunt.currentEnergyUsage.electricityUsed == true">
+          <label for="totalEmissionOutputRate">Total Emission Output Rate</label>
+          <div class="input-group">
+            <input name="totalEmissionOutputRate" type="number" step="any" class="form-control"
+              id="totalEmissionOutputRate" (input)="save()" 
+              [(ngModel)]="treasureHunt.currentEnergyUsage.electricityCO2OutputRate">
+            <span class="input-group-addon units small">kg CO<sub>2</sub>/kWh</span>
+          </div>
+        </div>
       </div>
     </div>
     <div class="d-flex">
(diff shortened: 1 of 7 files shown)


src/app/treasure-hunt/operation-costs/operation-costs.component.html
```diff
@@ -269,4 +281,34 @@ <h3>Annual Utility Usage and Costs</h3>
...
+        <button class="close" aria-label="Close" (click)="hideZipCodeModal()">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+      <div #modalBody class="modal-body p-5">
+        <app-assessment-co2-savings
+          [co2SavingsData]="co2SavingsData"
+          [isFormDisabled]="false"
+          [inBaseline]="true"
+          (emitUpdateCo2SavingsData)="updatePsatCo2SavingsData($event)">
```
nbintertech, line 300: Shouldn't say 'psat'

src/app/treasure-hunt/operation-costs/operation-costs.component.ts
```diff
@@ -37,12 +53,28 @@ export class OperationCostsComponent implements OnInit {
       this.saveSettings();
     }
     this.treasureHuntSub.unsubscribe();
+    this.zipCodeModalSub.unsubscribe();
+  }
+
+  ngAfterViewInit() {
+    this.zipCodeModalSub = this.zipCodeModal.onShown.subscribe(() => {
```
nbintertech, line 60: This should be in or called from in ngOnInit()

nbintertech (changes requested):
Structure and implementation in general look good. Other than my comments, we need a few more changes:
- Try the modal at one width size smaller
- I don't think we need three action buttons - should just be 'save' or 'apply' 
- Let's get a header on that modal to give some context for the user - ask Kristina. Could just say 'Egrid zipcode lookup' or whatever we're using in the header of psat/phast, 'CO2 Emissions' 
- Don't need a default for zipcode on a new TH
- Default zipcode isn't coming from global settings

src/app/treasure-hunt/operation-costs/operation-costs.component.ts
rhernandez-intertech: Everywhere else in the app when a modal subscription is called its called in ngAfterViewInit()

nbintertech (approved):

Issue #5159, opened 2021-12-07
Add CO2 savings fields for Other Fuels in TH


rhernandez-intertech:
Three fields will be added: Energy Source, Fuel Type, and Total Emission Output Rate
The values will come from the CO2 Fuel Savings Fuels database and will include any custom fuels the user has added. If the energy source and fuel type are know the output rate will be filled in with the value from the database, if the fuel type is unknow the user will have to manually enter the output rate. 
There will also be a link to a modal to calculate the Net CO2 emission rate from Mixed Fuels (similar to the net fuel cost modal in PHAST) 
 ![image.png](https://images.zenhubusercontent.com/60b7bed2e39f152beecc60b4/ee17c081-7338-4737-bd5b-9a33132cba0b)
![image.png](https://images.zenhubusercontent.com/60b7bed2e39f152beecc60b4/e42a5b7a-cb07-4c9d-b3e7-c95a23c13b5f)
![image.png](https://images.zenhubusercontent.com/60b7bed2e39f152beecc60b4/9f97c4e9-91a7-4970-860c-5b6cfedb16a9)

PR #5174, opened 2021-12-15
Issue 5159: Add CO2 savings fields for Other Fuels in TH

rhernandez-intertech:
connects #5159


src/app/calculator/utilities/co2-savings/co2-savings.service.ts
```diff
@@ -61,4 +61,5 @@ export interface Co2SavingsData {
   userEnteredBaselineEmissions?: boolean;
   userEnteredModificationEmissions?: boolean;
   zipcode?: string,
+  usage?: number
```
nbintertech, line 64: Use descriptive variable naming

src/app/treasure-hunt/operation-costs/operation-costs.component.ts
```diff
@@ -278,5 +295,77 @@ export class OperationCostsComponent implements OnInit {
...
+    this.treasureHuntService.modalOpen.next(true);
+    this.mixedCO2EmissionsModal.show();
+  }
+
+  hideMixedCO2EmissionsModal() {
+    this.treasureHuntService.modalOpen.next(false);
+    this.mixedCO2EmissionsModal.hide();
+  }
+
+  applyMixedCO2EmissionsModalData(mixedOutputRate: number) {
```
nbintertech, line 353: These two 'apply__' methods do different things but are named similarly. Use better descriptive naming - make sure the event emitters from the modal make sense with whatever changes you make

nbintertech (changes requested):
Double check:
- percent usage is calculating cumulative results in the modal the way it's supposed to with multiple fuels.
- Steam and hot water and natural gas are still in the lists for Other Fuels and their fuel type doesn't populate

nbintertech (approved):

Issue #4926, opened 2021-10-04
Add CO2 Emissions Results 


nbintertech:

koay9f:
small label thing:
need the "/yr" in the units, and for savings the (unit) should come after the word "Savings" like the others
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/4fb8b5c2-7195-45cd-b5cb-b0653026320c)

PR #5208, opened 2022-01-05
Issue 4926

nbintertech:
connects #4926 

Add co2 emissions to waste water


rmroot (changes requested):
Looks good, only thing.. form fields are not disabled when first going to explore opps page. Should probably fix that

image

rmroot (changes requested):
Looks good, electricity cost still not disabled on init in modify conditions

rmroot (approved):

Issue #4925, opened 2021-10-04
Add CO2 Emissions Results


nbintertech:

koay9f:
Total CO2 = Electrical CO2 + Fuel CO2
Electrical CO2 = site import power {kW}* operating hours {hrs} * electrical emissions factor {kg CO2/MWh} * convert(kWh to MWh)
Remember site import power is calculated in modifications (don't know if that matters)
Fuel CO2 = boiler fuel use (MMBTU/hr) * operating hours (hr) * emissions factor (kg CO2/MMBTU)
Here is one place you can find energy numbers
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/16415c34-3872-4390-9f26-3fcd615daa11)

Places to put the CO2 numbers
1 - Electricity CO2
2 - Fuel CO2
3 - Total CO2
A - (same place as 3) alternative where you just have 1, 2 & 3 with a CO2 bold header and Electricity/Fuel/Total as the subheader

Also...
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/185f629f-89fd-4eb8-ac5f-79f7d5f8a1f0)

While you're messing with the steam report, pls add the "Total Steam Generated" and "Total Operating Cost"  under a "Steam System Summary" header at the top.  This would be al alternative place for "3"

koay9f:
NOTE: for the above if it says "Annual ..." don't need the "/yr"  - #5248

PR #5210, opened 2022-01-05
Issue 4925

nbintertech:
connects #4925 

add co2 savings to steam

### DIFF
--- a/src/app/ssmt/operations/operations.component.html
+++ b/src/app/ssmt/operations/operations.component.html
@@ -10,57 +10,65 @@
         [inSetup]="inSetup" [idString]="idString" [isBaseline]="isBaseline"></app-operating-costs>
     </div>
 
-  <div class="pt-2 ml-2 mr-2">
-    
-  <label class="group-label w-100">Carbon Emissions From Fuel</label>
-  <div class="form-group">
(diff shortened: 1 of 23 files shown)


src/app/ssmt/operations/operations.component.html
```diff
@@ -9,5 +9,58 @@ <h3>Operating Conditions</h3>
       <app-operating-costs [(form)]="operationsForm" [selected]="selected" [settings]="settings" (emitSave)="save()"
         [inSetup]="inSetup" [idString]="idString" [isBaseline]="isBaseline"></app-operating-costs>
     </div>
+
+  <div class="pt-2 ml-2 mr-2">
+    
+  <label class="group-label w-100">Carbon Emissions From Fuel</label>
+  <div class="form-group">
+    <label for="{{'energyType'+idString}}">Energy Type</label>
```
rmroot, line 17: Not sure the "Carbon Emissions From Fuel" portion is correct. If "Electricity" is selected for "Energy Type" all the fuel selections are still there.

rmroot (changes requested):
1. Emissions results didn't get added to the results panel
2. Same disable field issues as waste water in modify conditions
3. Probably want to consider moving "Implementation Cost" below emissions so things lineup correctly..?
image

Issue #5119, opened 2021-11-18
Motor Inventory: Add power factor calculators throughout catalog 


AlexBotts:
Add the power factor calculator to the 4 power factor fields under "load characteristics" in the catalog 
![image.png](https://images.zenhubusercontent.com/5e4ffc1aa541aecb7b4865d7/8f8ba682-ad2b-4bd6-b89b-ac18f03df7b2)

nbintertech:
Implemented using the % efficiency at load from same section. Please confirm this is correct @AlexBotts

koay9f:
The buttons make numbers, but they are not the correct numbers

I think that you are passing the full load current into the PF calculator when you need to be passing the current at load. 
Look at the Motor Performance calculator for how it calculates current and PF at load. 
It probably uses "MotorCurrent.cpp" `MotorCurrent::calculateCurrent`

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/e380f666-f92d-4b55-9ac7-a03d6c204947)

PR #5244, opened 2022-01-18
Issue 5119 - Add motor power factor calculator to  catalog

nbintertech:
connects #5119


rhernandez-intertech (changes requested):
I think that the little question mark should be added after each power factor calculator, just like each estimate efficiency calc has the little '?' button that links to the help text.

rhernandez-intertech (approved):

Issue #5130, opened 2021-11-23
New FSAT Tab error flicker


nbintertech:
When opening a new FSAT the operations tab is marked with an error (red) - once I click the tab it goes away. That component is probably not being passed an update fsat or it's a timing issue. Did we just look at this in pumps? Please check pumps also.

![newfsatbug.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/ce7a38aa-9f75-4d61-8f85-256221eed6ec)

rhernandez-intertech:
Pumps does not have this bug, it is only in Fans

PR #5246, opened 2022-01-18
Issue-5130: New FSAT Tab error flicker

rhernandez-intertech:
connects #5130 
pumps did not have this bug


nbintertech (changes requested):
This will prevent Co2 global settings from populating correctly. Shoot me a message before you pick this one up again

nbintertech (approved):

PR #5252, opened 2022-01-18
Issue 5186 - Data explorer small screens

ysintertech:
Moved Prase Data button into a higher div and added margins to the bottom of the page so the Parse Data button is more visible.


nbintertech (changes requested):
In both places... instead of an inline style try to do it with bootstrap (mb-4 or mb-5). Otherwise use a class defintiion in the css file if that doesn't work.

nbintertech (approved):

Issue #5250, opened 2022-01-18
Stack Loss Calculator


koay9f:
Need to update the stack loss calculator to match the flue gas calculator.
I think we just need to add Fuel Temp to Gas fuel type. Don't think it needs anything else.

Update standalone and modal.

koay9f:
Okay, apparently I didn't pay enough attention and we have 2 different fields titled "Ambient Air Temperature"
I think I remember we decided to set "Combustion Air Temp" and "Fuel Temp" equal and ask 1 field (Amibient Air), but then we added ambient temp to the calculations.

Two possible solutions, I'll let you decide which we have time for
1 - remove one of the ambient air temp fields and set all three variables passed to the suite (ambient, fuel, combustion air temps) equal and just have one field (default = 60 F)
2 - make a accoridian? window with ambient air, comubstion air, and fuel temps AND "Mositure in combustion air" all in there. Title window: "Detailed Analysis". Set defaults to 60 F, 60F, 60F, 0% respectivly. 
(60F is 15.6C for the metric examples/defaults

koay9f:
apparently it broke. :(

Error comes when you hit apply and nothing happens
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/81553ca4-9d1f-4d18-a215-42549fbe23ff)

Math looks right

PR #5314, opened 2022-02-01
Issue 5250: Stack Loss Calculator update 

rhernandez-intertech:
connects #5250


src/app/calculator/steam/stack-loss/stack-loss.component.ts
```diff
@@ -152,6 +153,7 @@ export class StackLossComponent implements OnInit {
       }
     }
     if (this.inModal == true) {
+      this.cd.detectChanges();
```
nbintertech, line 156: If you haven't already please confirm that the root cause of whatever problem requires cd.detectChanges() can't be solved another way.

nbintertech (changes requested):

nbintertech (approved):

Issue #5196, opened 2021-12-30
Compressed Air: Specific Power 


AlexBotts:
Add specific power for each compressor in the report. 
There are two "Specific Power" and "specific power at average load" 

Specific power = kW / CFM  * 100  => units will be "kW/100 CFM"

For the regular "specific power" use full load values (performance points), for "specific power at average load" use the average kW and CFM for the "on" hours. There will be a "specific power at average load" for each daytype. 

![image.png](https://images.zenhubusercontent.com/5e4ffc1aa541aecb7b4865d7/8216f4b2-265e-415f-98ff-55d52e528303)

AlexBotts:
- This math is changing for specific energy. It is still kW/cfm at full load but we need to add a correction for the power.

AlexBotts:
To display the specific energy at average load, rated specific energy, and isentropic efficiency (issue 5195) we need to construct a table that is dynamic with the daytypes. Table: 
![image.png](https://images.zenhubusercontent.com/5e4ffc1aa541aecb7b4865d7/8242a408-0332-43e6-b408-cc442d4ab2fa)

This table should be shown in two locations.
 In system basics>system profile> *new tab* Compressor summary. 
It also needs to be displayed on the new tab being added in the report (issue 5335)

![image.png](https://images.zenhubusercontent.com/5e4ffc1aa541aecb7b4865d7/dcffae05-443b-47d7-808d-63fd33dd0109)

Issue #5195, opened 2021-12-30
Compressed Air: Isentropic efficiency 


AlexBotts:
Add a line item in the report for each compressor for the isentropic efficiency. The specific power  at full load (in issue #5196) will need to be calculated first.

Isentropic Efficiency = (16.52 *((((Discharge pressure + 14.5)/14.5)^0.2857)-1))/Specific Power at Full load 

![image.png](https://images.zenhubusercontent.com/5e4ffc1aa541aecb7b4865d7/fa5a88e8-59eb-453e-bdf9-6a01ee64dac9)

AlexBotts:
Later for help text: from CAGI
www.youtube.com/watch?v=VLjexIHLPqM

koay9f:
Specific powers are good, location etc is good. but Isentropic Eff is off.  Maybe paraenthese in the wrong spot?
(power((115+14.5)/14.5, 0.2857) - 1) * 16.57/19.58
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/206117a3-1a0b-46df-8836-e247590f1c1f)

Issue #4889, opened 2021-09-23
Compressor Profile avg operating point


koay9f:
Find avg operating point for each compressor for each bl/mod
put point on compressor profile

AlexBotts:
Notes for me later: This is going to be the average during the "on" hours for each day type (this will be needed to find the operational specific energy or the operation intensity). The second part of the issue: I think we wanted to point out where on the performance profile the compressor is operating

koay9f:
Add dots to profile graph in system setup
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/c7b7c7e8-3068-4f4c-93a7-ed28f8fbf07b)
Maybe add check box like "graph all compresssors" for "Graph Average Operating Points" 
Would need to be disabled until complete "System Profile"

do not include the op points on the graph in Assessment.

koay9f:
Looks good. Needs the different shapes for each daytype. (Tom suggested in weekly). Also, mabye up the size a bit

PR #5363, opened 2022-02-21
Issue 5196: Compressed Air Specific Power & Isentropic efficiency

rhernandez-intertech:
connects #5196 #5195 #4889


src/app/compressed-air-assessment/compressor-summary-table/compressor-summary-table.component.html
```diff
@@ -0,0 +1,69 @@
...
+                    <th></th>
+                    <th>Day Type</th>
+                    <th class="bold" *ngFor="let compressor of compressorInventoryItems">
+                        {{compressor.name}}
+                    </th>
+                </tr>
+            </thead>
+            <tbody>
+                <tr>
+                    <th [attr.rowspan]="dayTypes.length + 1">
```
nbintertech, line 21: It looks like you can use compressorSummaries length here and just keep one source of truth/dimensions

src/app/compressed-air-assessment/compressed-air-assessment-results.service.ts
```diff
@@ -521,6 +521,46 @@ export class CompressedAirAssessmentResultsService {
...
+  calculateRatedSpecificPower(compressor: CompressorInventoryItem): number {
+    let ratedSpecificPower: number;
+    ratedSpecificPower = (compressor.nameplateData.totalPackageInputPower / compressor.nameplateData.fullLoadRatedCapacity) * 100;
+    ratedSpecificPower = this.convertUnitsService.roundVal(ratedSpecificPower, 4);
+    return ratedSpecificPower;
+  }
+
+  calculateRatedIsentropicEfficiency (compressor: CompressorInventoryItem, ratedSpecificPower: number): number {
+    let ratedIsentropicEfficiency: number;
+    ratedIsentropicEfficiency = (16.52 *((((compressor.nameplateData.fullLoadOperatingPressure + 14.5)/14.5)^0.2857)-1))/ratedSpecificPower;
```
nbintertech, line 559: Most important, double check with Alex whether there should be a conversion done in either of these two calculation methods. Pressures look like they go psia/barg, airflow acfm/m3, but don't take my word for it.

Also, if there was a formula for either of these methods it would be nice to add as a comment. Helpful in discerning what the numbers represent.

nbintertech (changes requested):

src/app/compressed-air-assessment/compressor-summary-table/compressor-summary-table.component.html
rhernandez-intertech: the row span needs to be `dayTypes.length + 1` because the number of day types determines how many rows that cell needs to span.

nbintertech (approved):

Issue #5216, opened 2022-01-06
Shared Help Text for CO2


nbintertech:
- Create a shared help text component (in the shared folder) 
- Make sure the shared help text component has all existing help fields for CO2 emissions (check phast and the co2 standalone as reference)
- Add the shared help component to each results/help panel that exists in Fans, pumps, WW, CA, PHAST, TH, steam, motor inventory. This will replace what's already there. 
- Verify that some of the fields with differing labels i.e. 'Coal carbon emissions output' still reference the 'total emissions output field'

PR #5384, opened 2022-02-28
Issue 5216 - Shared Help Text for CO2

ysintertech:
connects #5216


src/app/shared/co2-help-text/co2-help-text.component.html
```diff
@@ -0,0 +1,84 @@
+<div class="my-2"
+    *ngIf="currentField == 'totalEmissionOutputRate' || currentField == 'eGridSubregion' || currentField == 'zipcode'">
```
nbintertech, line 2: - This condition needs the other two fields energySource and fuelType added.
- Move the condition to the ts file. Instead of explicitly using an if, try declaring currentField as a string literal + union type with the 5 values. I suspect if it's any value other than those, it may fall through as undefined and give the functionality we want.
https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types

nbintertech (changes requested):
- Let's add this to TreasureHunt also
- Not sure what you're doing with the 'fan-operations' changes. Fixing a tab badge bug? Shoot me a msg.

nbintertech (approved):

Issue #5333, opened 2022-02-07
Remove/modify disclaimers


rmroot:
remove/modify disclaimer that says "this application is in beta..."

koay9f:
Also the one for survey monkey.

PR #5402, opened 2022-03-04
Issue-5333: Remove/modify disclaimers

rhernandez-intertech:
connects #5333


rmroot (dismissed):
Each of these have corresponding typescript that can be removed as well.

nbintertech (changes requested):
Do an app search on disableDisclaimer and disableSurveyMonkey - those properties can be removed also, plus a leftover method in WW.

nbintertech (approved):

Issue #5335, opened 2022-02-08
CA: Performance Profile graphs


koay9f:
Change label name of graph
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/419c3edf-7644-46da-8cfc-15e87e3e5b8a)
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/94b690e7-f601-4e32-9c46-23dfe0431e5d)

Add Performance profile graphs to report - BASELINE ONLY
Add section for them (need to have all the bells & whistles for printing and rollup printing)
Include the dot for the avg operation points

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/51fd5bed-a11a-468d-945a-f9fd5d336649)

This is more important than issue #5319

koay9f:
This is the smallest, dumbest complaint. There should be a space between "Performce Profile" and "(Compressor NAME)"
Everything else is great.
The Performance Profiles are not printing properly, but that may be part of a bigger issue with CA printing...

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/d153625f-0d02-4a0b-9751-893c52d9bf93)

koay9f:
Performance profiles is in the CA printing settings, but not the global print settings. 
Printing is a bit off still. A bigger header for the section would be nice. why does it have a scroll bar?
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/bdac4fcf-a025-4920-9982-85ed39938fde)

koay9f:
unfortunatly still looks the same :(

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/f3a7cb0d-9cbb-4475-80b8-35ee7c9e9d4c)

Interesting note - if I have the app on my smaller monitor when I click print, it looks pretty good
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/262c177c-3cea-4873-a6a2-fd67c982d585)

Is this something beyond our control?

koay9f:
The graph is still being cut off in the roll up report and the assessment report. 
I mentioned this in the CA report issue, but will also toss this one back (I forgot that that portion was assigned to you Rachel).

PR #5408, opened 2022-03-07
Issue 5335: CA: Performance Profile graphs

rhernandez-intertech:
connects #5335


nbintertech (changes requested):
Missing this space from issue comments, otherwise good. 
image

nbintertech (approved):

Issue #5371, opened 2022-02-22
Steam New assessment bug


nbintertech:
From @koay9f 
![steambug20220222.gif](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/1df3fe61-e082-4389-bcf1-24676b7de24a)

PR #5425, opened 2022-03-10
Issue 5371- Steam New Assessment Bug

ysintertech:
connects #5371

### DIFF
--- a/src/app/ssmt/header/header.service.ts
+++ b/src/app/ssmt/header/header.service.ts
@@ -295,11 +295,14 @@ export class HeaderService {
 
         if (ssmt.boilerInput.blowdownFlashed == false && pressureLevel === 'highPressure') {
           headerType = 'high';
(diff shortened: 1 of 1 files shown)


src/app/ssmt/header/header.service.ts
```diff
@@ -295,11 +295,14 @@ export class HeaderService {
 
         if (ssmt.boilerInput.blowdownFlashed == false && pressureLevel === 'highPressure') {
           headerType = 'high';
+          let hasHeaderInputs: boolean = ssmt.headerInput.highPressureHeader !== undefined && ssmt.headerInput.highPressure !== undefined && ssmt.headerInput.mediumPressureHeader !== undefined && ssmt.headerInput.mediumPressure !== undefined && ssmt.headerInput.lowPressureHeader !== undefined && ssmt.headerInput.lowPressure !== undefined;
```
nbintertech, line 298: Don't need to check the mediums. The block that's conditionally wrapped only tries to access high and low

nbintertech (changes requested):

nbintertech (approved):

Issue #5369, opened 2022-02-22
Handle Steam Modeler inputs crash


nbintertech:
Steam modeler is crashing on the attached assessment when there is preheat water. 

Debug steam.service.ts call at ln 317 to see error
[https://app.zenhub.com/files/80439269/c564b5db-625b-4b1b-bfe2-0f2f91d77d92/download](https://app.zenhub.com/files/80439269/c564b5db-625b-4b1b-bfe2-0f2f91d77d92/download)
![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/2d03c83b-3d65-4dfd-9de7-82cd14e681af)

We need to implement an error message along the lines of "Steam Properties cannot be calculated something something. Please check input values." 

Collab with Kristina to decide in which places we'd like to display the message and how.

PR #5437, opened 2022-03-11
Issue 5369: Handle Steam Modeler inputs crash

rhernandez-intertech:
connects #5369


nbintertech (changes requested):
This isn't working in all cases and needs more testing, BUT I would think about starting over with a different approach. We should be able to:

- emit an error message to a behavior subject when the catch() block catches an error on steam modeler
- return an empty output object that will prevent results breaking on undefineds 
- subscribe to the bs in ssmt.component.ts and display the toast message

Issue #5485, opened 2022-03-17
Additional EAF Validation


koay9f:
Need validation on if calculated exhaust gas losses are too big
Cannot have more losses than fuel input, need to ask Sachin what a good limit would be (Exhaust gas loss must be < 0.7*chemical heat delivered)
'Electricity Input entered results in an Exhaust Gas Loss (###) greater than total fuel input (###)'

No flue/exhaust losses - 478.88 kW of fuel used
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/3912476a-ed91-4c87-9f43-82508fd1cdae)

Reasonable losses - 478.88kW fuel used and of that 213.62  islost
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/c13368ec-c93a-4fa0-8fa0-2c27ff128660)

More losses than fuel input - 478.88kW fuel used and of that 13213.62 is lost ?!?!?
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/4b6cbf34-ed7e-48fb-933f-161d17184ef5)

PR #5515, opened 2022-03-28
Issue 5485: Additional EAF Validation

rhernandez-intertech:
connects #5485


src/app/phast/losses/losses-tabs/energy-input-tab/energy-input-tab.component.ts
```diff
@@ -113,6 +117,9 @@ export class EnergyInputTabComponent implements OnInit {
       return this.energyInputCompareService.compareAllLosses();
     }
   }
+  checkElectricityInputWarning(): boolean{
```
nbintertech, line 120: Just call the resultService method from the this.inputError assignment.

nbintertech (changes requested):
Model this solution after other places in the app that check multiple warnings. The electricityInputWarning can be of type string with default of null. Set the validation message in the check method.

nbintertech (approved):

Issue #5509, opened 2022-03-24
Profile Summary Graphs Styling


nbintertech:
1. Fix headers
2. Fix missing scrollbar

I believe these were resized and aligned correctly at some point before CA printing style changes were made. Either way we should clean these up. Shortcut -  try looking at commit history in vscode to see when/where it was changed and revert back to that style without changing print rendering (if it's accurate).
![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/5614294c-40f0-4f23-aeae-e815f2e9d684)

koay9f:
Scroll bar still missing

koay9f:
Scroll bar is there - 
but now need a bit of padding at the bottom (sorry to keep adding)

PR #5525, opened 2022-04-04
Issue 5509 - Profile Summary Graphs Styling

ysintertech:
connects #5509

### DIFF
--- a/src/app/compressed-air-assessment/system-profile-graphs/system-profile-graphs.component.html
+++ b/src/app/compressed-air-assessment/system-profile-graphs/system-profile-graphs.component.html
@@ -1,21 +1,21 @@
 <div class="print-container">
     <div class="print-avoid-break">
-        <label class="group-label bg-white">{{labelName}} System Capacity</label>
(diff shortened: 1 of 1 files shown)


src/app/compressed-air-assessment/system-profile-graphs/system-profile-graphs.component.html
```diff
@@ -1,21 +1,21 @@
 <div class="print-container">
     <div class="print-avoid-break">
-        <label class="group-label bg-white ml-2">{{labelName}} System Capacity</label>
+        <label class="group-label bg-white">{{labelName}} System Capacity</label>
```
nbintertech, line 3: The label text is against the border of the div and should also get a pl 2 or 3.

nbintertech (changes requested):

src/app/compressed-air-assessment/system-profile-graphs/system-profile-graphs.component.html
ysintertech: good catch!

Issue #5436, opened 2022-03-11
WW - Influent CBOD5 help text 


koay9f:
Update Helptext

Influent CBOD5 concentration (So) is the concentration of CBOD5 that enters the activated sludge reactor, after a primary clarifier. The growth limiting substrate in the Bio-Tiger model is CBOD5. While the influent CBOD5 concentration changes during the day and throughout the week, this should be the average CBOD5 concentration for the particular activated sludge process being evaluated at steady-state.

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/aa0c56df-0c56-4e64-b896-2ef842e1b0b9)

![diagram-So.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/cda5ece9-968c-47dc-a4b7-20642e57f474)

NOT SURE IF WANT TO DO THIS PART - give me a bit of time to confirm, if I don't get it then we'll make a new issue for it later
A blue link for So 
Fields: Influent CBOD5
Help: Influent CBOD5 concentration (So) is the concentration of CBOD5 before a primary clarifier. The growth limiting substrate in the Bio-Tiger model is CBOD5. While the influent CBOD5 concentration changes during the day and throughout the week, this should be the average CBOD5 concentration for the particular activated sludge process being evaluated at steady-state.
Units: same as Influent CBOD5 in main form (I tihnk mg/L)
Name for math (suggestion): influentCBODbefore

Field: Primary clarifier Efficiency
Help: Approximate efficiency of the primary clarifier for removing BOD.
Unit: percent
Name for math (suggestion): clarifierEfficiency

Math that replaces what's in regular form `Influent CBOD5`
= influentCBODbefore * (1 - clarifierEfficiency/100)

Check: influentCBODbefore = 1000 mg/L, clarifierEfficiency = 70%
1000 * (1-70/100) = 300

koay9f:
Validation:
influentCBODbefore - cannot be less than or equal zero
clarifierEfficiency - cannot be less than or equal zero, cannot be greater than 100

koay9f:
1 - the picture to be used is slightly different from the one in the Diagram.  It's attached above and it has a big yellow arrow on it :)
2- The math for the calculated Influent CBOD5 is wrong. I was an idiot. 
```Influent CBOD5 = influentCBODbefore * (clarifierEfficiency/100)```
250 * (80/100) = 200
[...]

PR #5526, opened 2022-04-07
Issue 5436 - WW Influent CBOD5 help text

ysintertech:
connects #5436

### DIFF
--- a/src/app/shared/models/waste-water.ts
+++ b/src/app/shared/models/waste-water.ts
@@ -41,6 +41,9 @@ export interface WasteWaterData {
 export interface ActivatedSludgeData {
     Temperature: number,
     So: number,
+    isUserDefinedSo: boolean,
+    influentCBODBefore : number,
+    clarifierEfficiency: number,
     Volume: number,
     FlowRate: number,
     InertVSS: number,
@@ -151,6 +154,9 @@ export interface CalculationsTableRow {
 export interface WasteWaterTreatmentInputData {
     Temperature: number,
     So: number,
+    isUserDefinedSo: boolean,
+    influentCBODBefore: number,
+    clarifierEfficiency: number,
     Volume: number,
     FlowRate: number,
     InertVSS: number,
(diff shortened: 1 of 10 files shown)


nbintertech:
Github isn't allowing file upload right now, msg me if you need reference...

The validation error border style needs some work and the result should show a dash if it has no value. The tab button style has been changed so it's completely white when active - that may have been introduced elsewhere, but check it out please.

nbintertech:
Github isn't allowing file upload right now, msg me if you need reference...

The validation error border style needs some work and the result should show a dash if it has no value. The tab button style has been changed so it's completely white when active - that may have been introduced elsewhere, but check it out please.

src/app/shared/models/waste-water.ts
```diff
@@ -41,6 +41,9 @@ export interface WasteWaterData {
 export interface ActivatedSludgeData {
     Temperature: number,
     So: number,
+    isUserDefinedSo: boolean,
```
nbintertech, line 44: Test these fields with an old assessment, and update update-data.service if necessary. isUserDefinedSo may come up undefined/false and show invalid even if they had a value previously

nbintertech (changes requested):

ysintertech:
created separate issue for the whiteout styling bug #5536

src/app/waste-water/activated-sludge-form/activated-sludge-form.component.html
```diff
@@ -26,10 +26,54 @@ <h3>Activated Sludge Data</h3>
             </span>
         </div>
         <!-- So -->
+
+        <div class="form-group" *ngIf="!this.form.controls.isUserDefinedSo.value"
+        [ngClass]="{'indicate-different': activatedSludgeDifferent.influentCBODBefore, 'invalid': form.controls.influentCBODBefore.invalid}">
+            <label>Influent CBOD<sub>5</sub> Before</label>
```
nbintertech, line 32: The two new fields still don't use disabled styling (greyed out) when they aren't selected. Use the [readonly]="!selected" like the other fields on this form

nbintertech (changes requested):

nbintertech (approved):

PR #5530, opened 2022-04-12
Issue 5511: TH Project Tracking enhancements

rhernandez-intertech:
connects #5511 
adding two new npm packages: exceljs & file-saver
updating "@types/node" to 14.14.33
an npm install will be needed


nbintertech (changes requested):
The user can't (or I couldn't) see the total values until enable editing was clicked. Might be worth checking if there is a way around that. I wasn't able to see the better plants image either, if that was intended

Have a look at this https://angular.io/guide/build#configuring-commonjs-dependencies

Also the @types/node issue wasn't introduced by you. I'll have a small PR for bumping that package.

nbintertech (approved):

Issue #4966, opened 2021-10-07
Update IndexedDB implementation


nbintertech:
Research in VERIFI. This may help us in the web version

check out ngx-indexed-db to clean up calls to index db.
![image (1).png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/dcf9e6ea-e14a-4325-aa83-eaaed4b9cb85)

nbintertech:
Use this issue to prototype changes in one area - 5376 to complete remaining changes

nbintertech:
Required changes for web version on update-wasm-db

PR #5539, opened 2022-04-18
Issue 4966

nbintertech:
connects #4966 

- Update indexed-db to use wrapper service ngx-indexed-db
- Updates rxjs version
- update-wasm-db has changes current with this branch and develop


src/app/calculator/furnaces/energy-use/energy-use.component.ts
```diff
@@ -160,26 +161,6 @@ export class EnergyUseComponent implements OnInit {
...
-        this.indexedDbService.putCalculator(this.calculator).then(() => {
-          this.calculatorDbService.setAll();
-        });
-      } else {
-        this.saving = true;
-        this.calculator.assessmentId = this.assessment.id;
-        this.indexedDbService.addCalculator(this.calculator).then((result) => {
-          this.calculatorDbService.setAll().then(() => {
-            this.calculator.id = result;
-            this.calcExists = true;
```
rmroot, line 175: I think the `calcExists` logic of being turned to `true` when a new calculator is added is being lost here

src/app/calculator/compressed-air/air-flow-conversion/air-flow-conversion.component.ts
```diff
@@ -102,21 +102,18 @@ export class AirFlowConversionComponent implements OnInit {
     return tmpCalculator;
   }
 
-  saveAssessmentCalculator(){
+  async saveAssessmentCalculator(){
```
rmroot, line 105: any reason saveAssessmentCalculator is still being used in this calculator when the service method is being used in the others?

src/app/calculator/pumps/head-tool/head-tool.component.ts
```diff
@@ -145,31 +146,29 @@ export class HeadToolComponent implements OnInit {
...
         this.calculator.headTool = this.headToolService.getHeadToolFromForm(this.headToolForm);
         this.calculator.headToolSuction = this.headToolService.getHeadToolSuctionFromForm(this.headToolSuctionForm);
         this.calculator.headToolType = this.headToolType;
-        this.indexedDbService.putCalculator(this.calculator).then(() => {
-          this.calculatorDbService.setAll().then(() => {
-            this.closeTool();
-          });
-        });
+
+        let updatedCalculators: Calculator[] = await firstValueFrom(this.calculatorDbService.updateWithObservable(this.calculator))
```
rmroot, line 157: not using save method in service again, if done with a purpose no problem

rmroot (changes requested):
I tagged a couple spots but I don't want to go through every calculator file so two things: 

1. There is a saveAssessmentCalculator method that is being used in some calculators and not in others. I'm not sure if this is done intentionally or not but should double check that
2. The previous saveCalculator methods set a boolean of "calcExists" that is checked on changes to the calculator to know if it can save. Currently that boolean is never set to true when a new calculator is added so I don't think on changes to a new calculator it will be saved. The second time into a calculator in an assessment I think it may save.


Gonna leave it at that for know and will review again after you have looked at these two things

src/app/calculator/compressed-air/air-leak/air-leak.component.ts
```diff
@@ -91,7 +92,7 @@ export class AirLeakComponent implements OnInit, AfterViewInit {
     this.airLeakService.calculate(this.settings);
     if (this.assessmentCalculator) {
       this.assessmentCalculator.airLeakInput = this.airLeakService.airLeakInput.getValue();;
-      this.saveAssessmentCalculator();
+      this.calculatorDbService.saveAssessmentCalculator(this.assessment, this.assessmentCalculator);
```
rmroot, line 95: For this to works properly `calculate()` will need to be marked `async` and `await` will have to be added to the `saveAssessmentCalculator()` call

rmroot: Investigate lag on changes to calculator when this is implemented.

rmroot (changes requested):

rmroot (changes requested):
Looks good. Bleed test has a couple chunks of commented code

rmroot (approved):

Issue #5536, opened 2022-04-14
WW Assessment tab styling


ysintertech:
While in the assessment tab, if anything from the baseline is invalid the assessment banner is whited out

![Screenshot_92.jpg](https://images.zenhubusercontent.com/61e1a7fed1fded7035402cba/0461a484-69a2-430d-9002-9d5c2bd3bbc0)

PR #5541, opened 2022-04-21
Issue 5536 - WW Assessment tab styling

ysintertech:
connects #5536


nbintertech (changes requested):
There's no inputs or results to interpret on the diagram - so that can still be shown if invalid.

nbintertech (approved):

Issue #5310, opened 2022-01-31
CA Assessment and Rollup printing


nbintertech:
From @koay9f 
- Add Compressed Air Rollup Summary to global settings and modal
- don't think compressed air is hooked up to the print settings (I tried to just print "results" and "payback opportunities" and I got everything)
- Check CA assessment printing

ysintertech:
Reports seem to print out the current tab you're looking at whether you've chosen to include them in the report or not, in both the assessment page and the rollup report. This seems to affect multiple reports and a new issue has been created specifically for that problem #5329

koay9f:
The system profiles are not printing well (there was a seperate issue for this so maybe remove them for now)
Each section should start  on a new page.

koay9f:
Need to tidy up a bit more
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/7cb24c32-b923-41fd-be16-c88ce6a0b325)

System Profiles:
Remove the option to print as well
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/0e131dbf-4c39-4d53-a8e8-418e7f5a7fad)

Nothing is between the system profiles and payback period
and Payback period has a header but nothing in it
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/0749d1a3-587f-4827-bb47-b428772fa192)

koay9f:
Perf. Profiles graph is cutting off but otherwise looking good on individual print
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/61885b93-bed9-4e5b-9003-055b7b78d75a)

seems like some of the work to make the print reports look good don't transfer to the rollup report.
* Inventory Data is split on two pages
* System profile graph cut off
* all graphs cut off (sys cap, air flow cap, sys power, compressor power, perf. profiles)
* some scroll bars are back
[https://app.zenhub.com/files/80439269/848ebe31-102b-4f56-9440-fc77db64ab07/download](https://app.zenhub.com/files/80439269/848ebe31-102b-4f56-9440-fc77db64ab07/download)

ysintertech:
System profiles table being cut off will be handled in its own issue #5557

PR #5558, opened 2022-04-28
Issue 5310 - CA Assessment and Rollup Printing

ysintertech:
connects #5310


ysintertech:
The system profiles table _**is**_ currently being cut off, however, that will be handled in its own separate issue. This PR is merely fixing all of the issues with System Profiles graphs, Performance Profiles graph, removes scroll bars from print view, and prevents inventory data from being split in print view.

src/app/compressed-air-assessment/system-profile-graphs/system-profile-graphs.component.ts
```diff
@@ -346,10 +348,10 @@ export class SystemProfileGraphsComponent implements OnInit {
         }
         traceData.push(trace);
       });
-      let xRangeMax: number = this.profileSummary[0].profileSummaryData.length > 1 ? 24 : 1;
-      var layout = this.getLayout("Compressor Capacity (%)", [1, xRangeMax], [0, 105], '%');
+      let xRangeMax: number = this.profileSummary[0].profileSummaryData.length > 1 ? 25 : 1;
```
nbintertech, line 351: We shouldn't change any of the range values for the charts unless approved by Kristina/CA team. I think we might have tried to add a padded axis like this when it was created, and they preferred it without.

Another option: in non-print view display the charts with the original axis. When in print view assign the print view values from an object.

nbintertech (changes requested):

nbintertech (approved):

Issue #5543, opened 2022-04-21
 TH Report PPT - Add a slide for each opportunity 


rhernandez-intertech:
continuing the work of issue-2990, 
Adding a slide for each opportunity to the TH report ppt
![image.png](https://images.zenhubusercontent.com/60b7bed2e39f152beecc60b4/6462a1eb-a695-47a4-9679-c60d6017c120)

Issue #2990, opened 2019-05-03
TH - Export report as ppt


rmroot:
See work effort for exporting report as powerpoint.

If doable, implement

koay9f:
Basically one graph or table per slide
Maybe Title slide = Facility Name + "Treasure Hunt"
one slide with table with Annual Energy & cost

shivasaurus:
The work I completed on this is on the branch "issue-2990". Currently it is in a very partial state. The pptxgenjs library is used to create the export. I was able to get data tables to export using the tableToSlides() function of that library. The graphs, however, need to be saved to images when they are drawn/updated in Plotly. Then add the graphs as images to the PowerPoint export. Examples of this can be seen with the Team Summary Pie Chart. The sizing may need to be adjusted based on the image.

nbintertech:
Analyze where this branch is at and lets talk about how/when to move forward

rhernandez-intertech:
this issue encompasses the initial implementation of exporting TH report to PPT.
The ppt for this issue includes the title slide, executive summary tables, charts and graphs and payback details table, chart and graph

PR #5559, opened 2022-04-28
Issue 2990 Export report as ppt

rhernandez-intertech:
branch name is issue-2990-part2
connects #2990 #5543


src/app/treasure-hunt/treasure-hunt-report/treasure-hunt-report.component.ts
```diff
@@ -166,4 +181,45 @@ export class TreasureHuntReportComponent implements OnInit {
...
+
+  present() {
+    this.presenting = true;
+    this.cd.detectChanges();
+    if (!this.fileName) {
+      this.fileName = this.getFileName();
+    }
+    let facilityInfo: FacilityInfo;
+    let settings = this.settingsDbService.getByDirectoryId(this.assessment.directoryId);
+    if (settings) {
```
nbintertech, line 207: This area can be tightened up:
- currently, this logic and createPPT() both check facilityInfo exists. We can just pass settings.facilityInfo into createPPT and do ```if (facilityInfo && facilityInfo.facilityName)```  in that method
- Do we need to check settings exist? Do we need to give feedback if they click to export and there are no settings?

nbintertech, line 207: This area can be tightened up:
- currently, this logic and createPPT() both check facilityInfo exists. We can just pass settings.facilityInfo into createPPT and do ```if (facilityInfo && facilityInfo.facilityName)```  in that method
- Do we need to check settings exist? Do we need to give feedback if they click to export and there are no settings?

nbintertech (changes requested):
We're still passing 'presenting'/'exportToPPt' as Input() down through a number of components. We should be able to get by with one boolean in th report that displays exec-summary and opp-payback, and one behaviorsubject that the table components can subscribe to.

Reason: we don't need to pass state through components that aren't using it, and it's much easier/cleaner not having to follow a cascade of Input bindings. Instead we allow the component (table.components) that IS using the state to handle it themselves.

src/app/treasure-hunt/treasure-hunt-report/treasure-hunt-ppt.service.ts
```diff
@@ -0,0 +1,836 @@
...
+
+      newSlide.addTable(rows, { x: 1.14, y: 5.2, w: 11.05, colW: [1.5, 1.5, 0.8, 1.25, 1.25, 1.25, 1.25, 1.25, 1], color: "1D428A", fontSize: 12, fontFace: 'Arial (Body)', border: { type: "solid", color: '1D428A' }, fill: { color: 'BDEEFF' }, align: 'left', valign: 'middle' });
+      counter++;
+    });
+
+    return pptx;
+  }
+
+  roundValToString(num: number): string {
+    return Number(num.toFixed(2)).toLocaleString('en-US');
```
nbintertech, line 301: toString() should work for you here. Read the doc on toLocaleString

src/app/treasure-hunt/treasure-hunt-report/treasure-hunt-report.component.ts
```diff
@@ -166,4 +176,43 @@ export class TreasureHuntReportComponent implements OnInit {
...
+  }
+
+  getFileName(): string {
+    if (!this.fileName) {
+      this.fileName = this.getCurrentDate() + ' - Treasure Hunt Report';
+    }
+    return this.fileName;
+  }
+
+  present() {
```
nbintertech, line 201: The local variable facilityName reads a little confusing. It's the title 'Treasure Hunt Report", then it's the facility name + the title if we have a facility name. Then taken as the argument named facilityName (but it's the main slide title?) in the method.

You can remove the facilityName param and perform the logic from lines 207/208 inside the createPPT method. Settings is already being passed

nbintertech (changes requested):

src/app/treasure-hunt/treasure-hunt-report/treasure-hunt-ppt.service.ts
rhernandez-intertech: The toLocaleString('en-US') automatically adds comma formatting

nbintertech (approved):

Issue #4603, opened 2021-06-29
Psychrometric graph


koay9f:
Tidy Version
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/673a02c5-886e-4410-998f-7e5fcd53fa8a)

color coded version
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/b8c64811-c870-408d-8fd3-3839c9279cae)

bottom X-axis = "Dry Bulb Temperature" (unit = user choice)
Y-axis = "Humidity Ratio" (no unit)
ideally: 
"Top x-axis" = "Wet Bulb Temperature"
and labels for Relative Humidity

### _**Curved lines (blue)**_
Make lines that are at constant relative humidities 
Constants: User set: Barometric Pressure, Specific Heat Ratio, Gas Type, static pressure
Constant: Hard coded: Method == Rel Humidity
List 2: Dry Bulb = sequence(35 , 130, by 5) ? May want to increase the increments for better resolution
List 1: Relative Humidity = sequence(0, 100, by 5) (may want to change to by 10 if too crowded)
y1 = Humidity Ratio (one line for each Relative Humidity)
This will create a list of values to make lines: x = Dry Bulb, y = Humidity Ratio, one line for each Relative Humidity
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/1a275af5-b04b-4342-92a4-d8d7b1f5efaf)

### _**Straight lines (red)**_
Constants: User set: Barometric Pressure, Specific Heat Ratio, Gas Type, static pressure
Constant: Hard coded: Method == Wet Bulb
List 2: Dry Bulb = sequence(35 , 130, by 5) ? {same as above} May want to increase the increments for better resolution
List 1: Wet Bulb = sequence(35 , 130, by 5) (may want to change to by 10 if too crowded)
y2 = Humidity Ratio (one line for each Dry Bulb)
This will create a list of values to make lines: x = Dry Bulb, y = Humidity Ratio, one line for each Wet bulb temp
NOTE: if result "Relative Humidity" > 1 then do not display. This may be solved by filtering Wet Bulb list so that Wet Bulb <= Dry Bulb
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/f549e2b2-408c-4e70-9668-e735b484b75c)

nbintertech:
[...]

Issue #5568, opened 2022-05-05
Psychrometric Graph - Vertical Grid


ysintertech:
Vertical grid lines should get cut off when they intersect with the highest curved line

Issue #5567, opened 2022-05-05
Psychrometric Graph - Colors


ysintertech:
Make hard-coded lines black/gray instead of blue/red

Make lines created by user point red/blue

PR #5592, opened 2022-05-10
Issue 4603 - Psychrometric Chart

ysintertech:
connects #4603


ysintertech:
issues #5567 and #5568 are in this PR as well since they needed to be included in the release

nbintertech (changes requested):
- Delete the entirety of charlottes old psychrometric chart component
- convertTemperatures is unused, move it to the convertUnitsService for a rainy day
- Remove any comments that aren't telling why something unexpected was done
- Double check if we still need isBuildingGraph in psychrometric service
- Generally check for unused vars and methods left from prototyping

nbintertech (approved):

Issue #5557, opened 2022-04-28
System Profiles Table cutting off in print view


ysintertech:
Need to split the table in half during print view or fit the entire table into the print view

rhernandez-intertech:
System Profiles table split for printing and printing Baseline Profile Summaries for all Day Types

PR #5593, opened 2022-05-10
Issue 5557: System Profiles Table cutting off in print view

rhernandez-intertech:
connects #5557 
System Profiles table split for printing and printing Baseline Profile Summaries for all Day Types


nbintertech (changes requested):
I think it would be more clean to split the chart segments so each starts on a new page, but clear with Kristina please.

Issue #5128, opened 2021-11-22
Insulation Upgrade


koay9f:
For both pipe and tank insulation

Big change: update fuel type to be "Energy Source Type" 
Options: Natural Gas, Other Fuel, *Electricity*, steam
Other things effected by this change
* Utility cost: units depend on choice for energy source (electricity should be kWh, others are MMBTU)
* Calculation options (should appear on electricity & steam lists)
* utility that is changed by the calculator (reports, treasure chest, etc.)

Heat pipes or chilled water pipes:
add a dropdown after Energy source type that is __NEED NAME__, options: "Chilled", "Heated"
if heated - no change
if chilled - change "Heat Loss" result to "Heat Gain", anything from the suite would need to be x (-1) {If only the heat losses are coming from the suite, then only they need x(-1), but if it is also costs or the savings, then they also need (-1)}
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/33a00785-940f-43a5-aedc-56e8bd7f4819)

AlexBotts:
This calculator is also currently hidden if there is not a natural gas usage, needs to be available for both electric and natural gas

koay9f:
Re: Alex's comment - that is what was meant by "Options: Natural Gas, Other Fuel, Electricity, steam"
It needs to be available for all of those

rhernandez-intertech:
Tanks are not behaving properly so they will not receive this upgrade until they are ready. Pipes are ready for the upgrade at this time. 
***Don't merge into Develop until tanks get their upgrade***

koay9f:
Pipes & Tanks
-  If Energy Source Type = Electricity - it is not getting the electricity default value

Tanks
If setup for chilled (ambient > tank temp) but heated/chilled = heated, {or setup for heated and h/c = chilled} baseline is negative (as should be) but mod is not.  as they should be using the same calculations, don't know why this is.  Low priority, but would be nice to understand

PR #5611, opened 2022-05-16
Issue 5128: Insulation Upgrade

rhernandez-intertech:
connects #5128


nbintertech (changes requested):
Functions as needed, but needs some code improvement - below is for both pipe and tank:

- Instead of handling it with template/ngIf conditions, the form and report components should have a variable that represents utility unit. Set that with a getEnergyUnitMethod based on the service. See wall calculator for example.
- utilityCost/type are setup weirdly redundant in the forms, and I'm guessing they were used as reference for heatedOrChilled. heatedOrChilled is already set on the form in the main calc component and does not need to be passed in or handled in ngOnChanges. It should behave like all the other form properties. Don't worry about changing utilityCost/type

src/app/calculator/steam/pipe-insulation-reduction/pipe-insulation-reduction-form/pipe-insulation-reduction-form.component.ts
```diff
@@ -26,6 +26,8 @@ export class PipeInsulationReductionFormComponent implements OnInit {
   @Input()
   utilityCost: number;
   @Input()
+  heatedOrChilled: number;
```
nbintertech, line 29: Remove

nbintertech (approved):
I was wrong about the second bullet point Re: utilityCost/type/heatOrchilled. Will push a commit to fix

Issue #5612, opened 2022-05-18
Fan Inlet & Outlet pressure modal not updating in Assessment


koay9f:
In explore opps the inlet outlet pressure modals don't have a "Save and Exit" and "Apply Data" doesn't work (doesn't update calcs from BL and doesn't save or apply). Seems to work in expert view
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/738c9064-9361-45ca-ae99-3b7e8b326a20)

rhernandez-intertech:
once the Inlet and Outlet pressure modals updated things properly, compressibility factor needs to be updated from Explore Ops or else a "Change after Check" error occurs when switching to Expert View. So that was addressed in this issue as well

PR #5623, opened 2022-05-23
Issue 5612: Fan Inlet & Outlet pressure modal not updating in Assessment

rhernandez-intertech:
connects #5612

### DIFF
--- a/src/app/fsat/calculate-pressures/calculate-inlet-pressure/calculate-inlet-pressure.component.ts
+++ b/src/app/fsat/calculate-pressures/calculate-inlet-pressure/calculate-inlet-pressure.component.ts
@@ -73,7 +73,9 @@ export class CalculateInletPressureComponent implements OnInit {
   }
 
   calculate() {
-    this.setInletVelocityPressure();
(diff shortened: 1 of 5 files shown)


nbintertech (changes requested):
- Both modal and the disabling of buttons are buggy. If data is entered, sometimes the button is disabled, sometimes not.
- I think more information is needed here with respect to how these should operate. There's some result validation going on but no messaging to the user - I couldn't produce it anyway.
Will upload screens when github starts working

nbintertech (changes requested):
- The Apply Data buttons are never disabling so we should get rid of the boolean and any helper method pertaining to that if that's the correct behavior
- I didn't encounter any ngchangedAfterChecked, nor anything related to compressibilityFactor within the explore opps form here. Remove anything related to that otherwise let me know if there's a way to reproduce that. Field data - compressibilityFactor isn't set properly on a new setup/assessment, but that should be a separate issue
- Min validation is broken. 
image

nbintertech (approved):

Issue #5595, opened 2022-05-11
Data Explorer seconds getting lost


koay9f:
Figure out something to help when seconds get lost b/c of bad data formatting (problem for csv, maybe xlsx)

Maybe something that lets the user say, well it is 5 second data and we just concencate things together?

rhernandez-intertech:
The seconds are getting "lost" when the data is being read from an .xlsx or .cvs file.
This is messing up how the data gets graphed.

PR #5635, opened 2022-05-26
Issue 5595: Data Explorer seconds getting lost

rhernandez-intertech:
connects #5595

### DIFF
--- a/src/app/log-tool/system-setup/clean-data/clean-data.component.ts
+++ b/src/app/log-tool/system-setup/clean-data/clean-data.component.ts
@@ -24,8 +24,7 @@ export class CleanDataComponent implements OnInit {
   editField: LogToolField;
   individualDataFromCsv: Array<IndividualDataFromCsv>;
   dateExistsForEachCsv: boolean;
-  intervalForSecondsSub: Subscription;
-  intervalForSeconds: number;
+  dataIntervalValidSub: Subscription;
   secondsIntervalOptions: Array<number> = [ undefined, 1, 2, 5, 15, 20, 30 ];
   constructor(private logToolService: LogToolService, private logToolDataService: LogToolDataService, private cd: ChangeDetectorRef,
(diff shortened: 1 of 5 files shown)


src/app/log-tool/system-setup/clean-data/clean-data.component.ts
```diff
@@ -19,9 +20,13 @@ export class CleanDataComponent implements OnInit {
   dataSubmitted: boolean = false;
   dataExists: boolean = false;
   showEditModal: boolean = false;
+  dataIntervalValid: boolean = false;
   editField: LogToolField;
   individualDataFromCsv: Array<IndividualDataFromCsv>;
   dateExistsForEachCsv: boolean;
+  intervalForSecondsSub: Subscription;
```
nbintertech, line 27: Can we get rid of the intervalForSeconds Sub and BS? Keep the intervalForSeconds component property and pass it to submitIndividualCSVData (make it a property of that object).

With the current setup, intervalForSeconds is set when the user makes a selection, and also set redundantly when that intervalForSeconds sub is emitted to.

src/app/log-tool/system-setup/clean-data/clean-data.component.ts
```diff
@@ -94,4 +106,9 @@ export class CleanDataComponent implements OnInit {
     // });
     this.cd.detectChanges();
   }
+
+  setIntervalForSeconds(){
+    this.logToolDataService.intervalForSeconds.next(this.intervalForSeconds);
+    this.cd.detectChanges();
+  }
```
nbintertech, line 113: Do we need this method? With ng template driven forms --> [(ngModel)]="intervalForSeconds" should bind to the component property when the user selects. Is there something else going on here requiring the method and detectChanges()?

nbintertech (changes requested):

nbintertech (approved):

Issue #5596, opened 2022-05-11
Psychrometric Calc Data validation


koay9f:
Can put some bad numbers in.  Doesn't cause crash but does make the graph weird and spits out insane results.

No good way to do it from inputs.  Basically if relative humidity, humidity ratio or degree of saturation are negative, the results are nonsense.  
Maybe look and see why Dew Point is "--"?

If rhumidity ratio < 0 - don't display results and display message
Message: `Data is producing a negative humidity ratio, please review data`

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/214ca4f3-3eff-483a-92a1-f930bf9ceba5)

nbintertech:
I found that the alpha used to calculate dewpoint is also NAN, resulting in the same for dewpoint. Looks like this goes back to negative numbers for the previously calculated saturationPressure and humidityRatio. Haven't dug any further.

```
alpha = std::log(absolutePressure * 0.4911541 * humidityRatio / (nMol + humidityRatio))
```

nMol = .62198
```
std::log(26.57 * 0.4911541 * -0.000420905 / (nMol + -0.000420905))
```

PR #5639, opened 2022-06-02
Issue 5596 - Psychrometric result validation

nbintertech:
#5596 

- Add results validation and hide results for psychrometric components


rhernandez-intertech (changes requested):
validation is not updating,
image

Altitude field not going away when reset and/or example btn clicked 
And copy table does not clear when example btn clicked, 
image

rhernandez-intertech (approved):

Issue #5652, opened 2022-06-06
Chiller Staging Calc improvements


koay9f:
* Defaults for "Design/Max Capacity Ratio" 
     - if chiller type = "Centrifugal" --> default = 0.92
     - if chiller type = "Screw" --> default = 1
* Disable "Design/Max Capacity Ratio" field if selected chiller type = "Screw" (and make sure value =1)
* Reset Data --> Chiller type = "Screw", Ratio = 0.92

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/9e16d0de-acd3-4557-b390-c6dcb801f248)

JL102:
Note: Waiting for word from Wei before doing the third bullet point.

JL102:
Testing note: Prior to fixes, switching to type = Screw and then clicking Generate Example or Reset Data would keep Motor Drive Type & Compressor Config Type disabled.

PR #5662, opened 2022-06-10
#5652 Chiller Staging/Performance calc improvement

JL102:
Waiting for word from Wei on whether we do want to switch the default cooler type from Centrifugal to Screw (third bullet point of #5652). This PR does not include that requested change.


nbintertech (changes requested):
There is sometimes a timing issue when using formbuilder to set forms. For both ts. files here you'll probably need to force Angular change detection (see uses in app of "ChangeDetectorRef" and "cd.detectChanges()"). Remove timeouts and comments.

Issue #5616, opened 2022-05-19
Handle CA CO2 input display when no emission rate


nbintertech:
![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/0bd3f602-ad99-48c4-acd5-0488661e9eed)

JL102:
reproduction notes
Step 1) Create a new Compresses Air assessment. 2) Fill out the assessment: you can enter 500 for total air storage in system information and set data from an existing compressor in inventory. 3) Go to the Report tab. 4) click print -> select all ->  print. Then the errors will pop up in DevTools console

PR #5664, opened 2022-06-10
#5616 Handle CA CO2 input display when no emission rate

JL102:


nbintertech (changes requested):
Looks good. Remove comments please

nbintertech (approved):

Issue #5629, opened 2022-05-24
Update assessments when sql db items deleted


rmroot:
If a phast assessment has a custom material selected and that material is deleted from the database, we need to update the assessments.

rhernandez-intertech:
Charge Material losses handled in this issue, remaining losses that use custom materials will be handled in separate issues

PR #5673, opened 2022-06-14
Issue 5629: Update assessments when sql db items deleted

rhernandez-intertech:
connects #5629 
Only implemented for Charge Material losses, will handle the rest of the losses in separate issues


nbintertech (changes requested):
This solution does the job for us but there are a couple things we can condense - mainly that we are already adding the material and asking the user if they want to redundantly add. We should just add it and refer them to the Custom Materials area if needed, or we should manage the data integrity another way so the dummy isn't added to the database.  Let's discuss before you continue.

nbintertech (approved):

Issue #5532, opened 2022-04-13
Custom Material Imports


nbintertech:
1. Clicking to import an exported set of custom materials does nothing
2. If the exported set of materials is named something other than the default w/date the import button is disabled 

![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/52ce1068-2c10-43a6-ac32-f92317853526)

PR #5682, opened 2022-06-15
#5532 Custom material imports

JL102:
connects #5532


nbintertech (changes requested):
The current changes fix issue point 1, but the behavior from the 2nd point still persists.

nbintertech:
Still able to reproduce the 2nd point. More specific repro steps --> use a custom filename that does not have the '.json' extension. If you want reference, an implementation that guards against this is already implemented in a few other areas, i.e. exporting an assessment.

JL102:
As discussed offline, the agreed-upon resolution to issue #2) will be to emphasize the file type in the export button ("Export as .json") to keep consistency with the export assessments screen. That way, users shouldn't be confused if they change the exported file extension and wonder why it's not importing.

I wanted to modularize the import/export modal dialog components so that the materials & assessment guys use shared code, but it would take more time than originally expected. We can put that into a new issue if desired.

nbintertech (approved):

Issue #5655, opened 2022-06-07
Help text for disabled fields


koay9f:
It would be nice if we could still have help text for disabled fields, so users would know why it is disabled
Currently can't do b/c help text is dependent on clicking. 
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/da50df8f-b459-411f-b564-cce2983365eb)

JL102:
Help text does not show in this case because it is tied to the inputs' `focus` events, which do not fire when an input is disabled.
Since the app is monstrous, and the way help text is shown is well established, the optimal solution would not be to modify the way help text is shown (for example, as an onClick handler for specific inputs/labels which might sometimes be disabled). Rather, we should add a small script which forces the `focus` event to fire in the case that it (or its label) is clicked while it is disabled.

koay9f:
Why is this still in needs review and not in needs verification? Is it in develop?

PR #5685, opened 2022-06-15
#5655: Help text for disabled fields

JL102:
connects #5655
Testing note: It's easier to test in the branch `cooling-and-weather-2`, because the first two tests disable a few inputs when you set cooler type to Screw. Set the PR to merge into develop, however, because it's unrelated to cooling-and-weather.


nbintertech:
This is probably the best solution for us being that there isn't a clean or simple way to ask Angular to do this. I'd like to test thoroughly before we merge, so don't wait up for this one.

Issue #5689, opened 2022-06-16
Alex TH Bug CA Reduction Import


rmroot:
Import opportunities with compressed air reduction into an existing TH assessment breaks app. Typo in method.

PR #5690, opened 2022-06-16
Th import CA reduction bug

rmroot:
connects #5689 

This is a breaking bug for TH

Importing opportunities with compressed air reduction breaks without compressed air pressure reduction because import method is incorrect.


rmroot:
terrible. closing

Issue #5648, opened 2022-06-03
Total Downloads CSV/XLSX


nbintertech:
We need to be able to run a standalone script (probably python/nodeJs) that can call the github API for download data of MEASUR/VERIFI and then compile a spreadsheet showing total downloads

```doing "total downloads" include everytime a .yml file gets pulled which is everytime the app checks for an update or everytime the app is opened```

JL102:
[downloads.csv](https://github.com/ORNL-AMO/AMO-Tools-Desktop/files/8922579/downloads.csv)
Current output titles: Name, Date, Windows, Mac .DMG, Mac .ZIP, Linux AppImage, Linux Tar.GZ, latest.yml, latest-mac.yml, latest-linux.yml
Any requests/changes before it's finalized?

koay9f:
looks good!

PR #5701, opened 2022-06-17
#5648 Download statistics

JL102:
connects #5648

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,12 +15,13 @@
 - To start the electron app: `npm run electron`
 
 ## Package
-- To build a package:`npm install && npm run ng-high-memory` and then: `npm run dist`
+- To build a package:`npm install --save-dev && npm run ng-high-memory` and then: `npm run dist`
 - The packages will be placed in `../output`
-- Example: `npm install && npm run build-prod && npm run dist` will make a production-build installer for Windows
+- Example: `npm install --save-dev && npm run build-prod && npm run dist` will make a production-build installer for Windows
 
 ## For Developers
 - To generate a new angular module: `npm run ng g module path/to/module`
 - To generate a new angular component: `npm run ng g component path/to/component`
-- For more information, see [the angular docs](https://docs.angularjs.org/guide/component)
+- For more information, see [the Angular docs](https://docs.angularjs.org/guide/component)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -15,12 +15,13 @@
 - To start the electron app: `npm run electron`
 
 ## Package
-- To build a package:`npm install && npm run ng-high-memory` and then: `npm run dist`
+- To build a package:`npm install --save-dev && npm run ng-high-memory` and then: `npm run dist`
 - The packages will be placed in `../output`
-- Example: `npm install && npm run build-prod && npm run dist` will make a production-build installer for Windows
+- Example: `npm install --save-dev && npm run build-prod && npm run dist` will make a production-build installer for Windows
```
rmroot, line 20: We don't want `--save-dev` in these instructions. That's for installing new packages to a project and saving them to the dependencies in package.json

package.json
```diff
@@ -118,14 +119,15 @@
     "@types/lodash": "4.14.100",
     "@types/node": "^14.18.12",
     "angular-plotly.js": "^4.0.4",
+    "axios": "^0.27.2",
     "codelyzer": "^6.0.0",
     "electron": "13.6.6",
     "electron-builder": "22.9.1",
     "electron-download-tf": "4.3.4",
     "electron-packager": "15.2.0",
-    "jasmine-core": "~3.6.0",
```
rmroot, line 126: Why are we removing `jasmine-core` here?

rmroot (changes requested):

scripts/download-stats.ts
```diff
@@ -0,0 +1,152 @@
+import axios, { AxiosError, AxiosResponse } from 'axios';
+import { Parser } from 'json2csv';
```
nbintertech, line 2: See if papaparse (already a dependency) works for these needs and remove json2csv.

nbintertech (changes requested):

package.json
JL102: The specific version caused installation problems and it's not being directly used. It's already a sub-dependency, so it'll be installed anyways.

nbintertech (approved):

rmroot (approved):

Issue #5717, opened 2022-06-22
Improve visual style & consistency of navigation


JL102:
1. Elements in the sidebar feel cramped & too close to the edge of the screen. VERIFI gives its sidebar elements plenty of space from the side of the screen.
1. Vertical line representing an open assessments folder is inconsistently placed compared to that of calculator categories.
1. Gradient behind the scroll bars is unappealing and inconsistent with the flat visual style of the rest of the application.
1. The outline of focused elements (when pressing tab) is the default orange, which clashes with the green.

PR #5718, opened 2022-06-22
#5717 Visual improvement to side bar

JL102:
connects #5717
Purely visual modification

### DIFF
--- a/src/app/app.component.css
+++ b/src/app/app.component.css
@@ -1870,3 +1870,7 @@ No Data Messaging
       margin: 0;
   }
 }
+
+:focus-visible {
+  outline: rgb(47, 230, 117) auto 1.5px;
+}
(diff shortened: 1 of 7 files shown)


nbintertech:
Closing until ok'd by Oakridge team. Feel free to re-open if that changes

JL102:
Reopened with permission from @koay9f

src/app/app.component.css
```diff
@@ -1870,3 +1870,7 @@ No Data Messaging
       margin: 0;
   }
 }
+
+:focus-visible {
```
nbintertech, line 1874: Not sure what part of the app promped this since this style is app-wide - was it solely in the directory/sidebar?. 1. Can we include this style only in the parts of the app where this is the most egregious and 2. can we just use the native browser focus color like what's seen on form fields?

nbintertech (changes requested):

nbintertech (approved):

Issue #5743, opened 2022-06-28
Dropdown chevrons (v and ^) point in the wrong direction on Custom Materials and Settings


JL102:
Collapsed dropdown menus should appear with a downward chevron ("v"). Sometimes expanded dropdown menus keep the downward chevron, but sometimes they get a bit of visual flourish by inverting the chevron ("^"). The "About" page does this correctly, but Custom Materials and Settings do it incorrectly.

Custom Materials start out expanded, so their chevrons should start out like "^":
![image.png](https://images.zenhubusercontent.com/629f63e243d33b2a6a3a59f2/5a6715fe-4562-4f5f-9458-a1a5ff8d0c95)

Settings start out collapsed, so their chevrons should start out like "v":
![image.png](https://images.zenhubusercontent.com/629f63e243d33b2a6a3a59f2/d5edab78-5de7-46a2-a6e9-5380db20e591)

rmroot:
If expanded then chevron should be down

If collapsed then chevron should be up

Looks like just the About page should be changed.

Issue #5742, opened 2022-06-28
Custom Materials: "+Add X" button disappears when the dropdown is clicked, even if no materials for that category exist


JL102:
When no materials for a category exist, either:
1. hide the chevron (v ^) entirely, or
2. make the chevron not change orientation when clicked, or
3. Keep the dropdown behaving as-is, but simply make the "+Add X" button not disappear when clicked

![image.png](https://images.zenhubusercontent.com/629f63e243d33b2a6a3a59f2/4a840c03-40fa-475c-84b7-e11d7081b5c3)

JL102:
Personal opinion: item 3) may be visually inconsistent/strange to users, so I believe either 1) or 2) would be the best solution.

JL102:
@nbintertech Which option is preferable?

nbintertech:
Top down priority:
- Add button should always be visible
- Collapsed chevron points 3:00, open chevron points 6:00 (See process heating calculator help text)
- We can hide the chevron if there are no materials

JL102:
@nbintertech WIP code has been uploaded to branch https://github.com/ORNL-AMO/AMO-Tools-Desktop/tree/issue-5743
the requested changes are non-trivial because the "Add Custom" button requires the `app-custom-x-materials` component to be visible for the modal dialog to open. The logic is a bit messy and requires some reworking to get it to work consistently.

PR #5754, opened 2022-06-30
Issue 5742 and 5743: Custom materials dropdown

JL102:
Connects #5742 and #5743

### DIFF
--- a/src/app/suiteDb/custom-materials/custom-materials.component.html
+++ b/src/app/suiteDb/custom-materials/custom-materials.component.html
@@ -30,13 +30,14 @@
       <div class="card-form">
         <label class="group-label general" (click)="toggleMaterial('flue')">Custom Gaseous Fuels
           <span class="fa fa-chevron-down pull-right m-1" *ngIf="!showFlueMaterials"></span>
-          <span class="fa fa-chevron-up pull-right m-1" *ngIf="showFlueMaterials"></span>
+          <span class="fa fa-chevron-up pull-right m-1" *ngIf="showFlueMaterials && hasFlueMaterials"></span>
         </label>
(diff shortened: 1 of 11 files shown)


rmroot (changes requested):
See comment on issue 5743:

If expanded then chevron should be down

If collapsed then chevron should be up

Looks like just the About page should be changed.

src/app/suiteDb/custom-materials/custom-materials.component.html
```diff
@@ -29,14 +29,15 @@
...
+          <span class="fa fa-chevron-down pull-right m-1" *ngIf="!showFlueMaterials"></span>
+          <span class="fa fa-chevron-up pull-right m-1" *ngIf="showFlueMaterials && hasFlueMaterials"></span>
         </label>
         <a class="click-link pull-right small" *ngIf="showFlueMaterials" (click)="toggleModal('flue')">+Add Custom Gaseous Fuel</a>
         <!-- <p class="help-text">View, create, edit your custom flue gas materials.</p> -->
       </div>
     </div>
-    <app-custom-flue-gas-materials *ngIf="showFlueMaterials" [settings]="settings" [showModal]="showFlueModal" [importing]="importing"></app-custom-flue-gas-materials>
+    <app-custom-flue-gas-materials *ngIf="showFlueMaterials" [settings]="settings" [showModal]="showFlueModal" [importing]="importing"
+      (emitNumMaterials)="hasFlueMaterials = ($event > 0)"></app-custom-flue-gas-materials>
```
rmroot, line 40: We don't want to do logic for setting variables in the html template. Create a function that is called in the .ts

rmroot (changes requested):

src/app/suiteDb/custom-materials/custom-materials.component.html
JL102: Fixed

rmroot (approved):

PR #5828, opened 2022-08-03
Issue 4644, 4893 End Uses, Airflow sankey, Power sankey

nbintertech:
Big PR so, potential opportunity areas:
- when/how dayTypeAirflowTotals are set.
- when/how dayTypeSetup leak rates and selected id's are set


rmroot (approved):
Still need to hide the delete day types button when end uses exist. Warning is showing properly.

Issue #5837, opened 2022-08-03
Can't delete day type if end uses exist


koay9f:
need to hide the day types delete button when end uses exist

Issue #5836, opened 2022-08-03
End uses valdiation


koay9f:
a validation message if the any of the end uses are invalid (to the chart),

nbintertech:
Also adding the same validation and hiding the sankey

PR #5844, opened 2022-08-04
Issue 5836, 5837 Validation additions, Restrict day type modifications

nbintertech:
#5836 #5837

### DIFF
--- a/src/app/compressed-air-assessment/end-uses/end-uses.service.ts
+++ b/src/app/compressed-air-assessment/end-uses/end-uses.service.ts
@@ -211,7 +211,7 @@ export class EndUsesService {
       }
     }
 
-    return {hasValidEndUses: !hasInvalidEndUse, endUseEnergyData};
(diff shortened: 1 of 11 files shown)


src/app/compressed-air-assessment/end-uses/end-uses.service.ts
```diff
@@ -189,7 +211,7 @@ export class EndUsesService {
       }
     }
 
-    return endUseEnergyData;
+    return {hasValidEndUses: !hasInvalidEndUse, endUseEnergyData};
```
rmroot, line 214: I don't think that this is correct. I think it should be:

`{
  hasValidEndUses: !hasInvalidEndUses, 
  endUseEnergyData: endUseEnergyData
}`

as is I think endUseEnergyData will always be undefined.

rmroot (changes requested):

rmroot (approved):

PR #5867, opened 2022-08-11
Issue 5793: modify PSAT system setup for phone/table UI

rhernandez-intertech:
connects #5793


nbintertech (changes requested):
IMPORTANT: before you continue let's go over this offline. Your applied styles are fine, but there's an easier way to do this work using CSS media queries that doesn't involve copying the markup.

1. We should be taking up100% width when possible. (change needed in results/help, report, assessment results) 
2. Add padding to the sub ‘Menu” chevron
3. Clicking the hamburger menu nav (to file explorer, home, or settings) takes us to the desired route but should also automatically collapse the sidebar, so that we see the route's view.
4. Scroll containers: In system setup and assessment, we should only see one scroll bar. Currently there is the main container scroll as well as the results/help scroll

Strong nice to haves:
1. System setup > Field Data > Pump head calc - in-assessment modal styles 
2. Click System-setup > assessment settings. Hovering over the numbered tabs is buggy with respect to white/gray background

nbintertech (changes requested):
I'm excluding any sankeys as well as the diagram for any general issues below.

Print is shown as a button instead of list item/link. Report table and other areas are still not taking up 100% width
image

image

Non-matching double scrolls show below
image

On modify-conditions load, the component that shows first (baseline) is not highlighted as selected (white background)
image


Modification tab doesn't have an active style applied
image

Double scroll bars on the details page
image

nbintertech:
If you're mocking the styles from responsive-calculators.css in the PSAT, you may not need 'scroll-item' classes. Similarly, shouldn't need to set the container height with ngStyle

nbintertech:
Regular PSAT view no longer has the grey background color for help/results
image

Issue #5890, opened 2022-08-17
Stack Loss and Pipe Insulation UI


nbintertech:
Apply small screen styles

koay9f:
Cannot scroll all the way to the bottom

PR #5891, opened 2022-08-17
Issue 5890 - small screen styles for stack loss and pipe insulation

nbintertech:
connects #5890


rhernandez-intertech (changes requested):
image
double header, please fix

PR #5920, opened 2022-08-25
Issue 5757: Chiller Staging TH integration

rhernandez-intertech:
connects #5757


src/app/treasure-hunt/find-treasure/find-treasure.component.html
```diff
@@ -355,6 +355,21 @@ <h1 class="p-2 lead text-center mb-0">
     </div>
 
 
+
+    <!--chiller staging-->
+    <div *ngIf="(displayCalculatorType == 'All' || displayCalculatorType == 'Electricity') && treasureHunt.currentEnergyUsage.electricityUsed == true"
```
nbintertech, line 360: If we want all the process cooling cards together this is fine. Otherwise there's already a group for the electricity only calculators at line 53.

nbintertech (approved):

PR #5930, opened 2022-08-31
Issue 5866: Apply Mobile UI to Steam calculators

rhernandez-intertech:
connects #5866


nbintertech (changes requested):
Saturated Properties details tab isn't functioning

nbintertech (approved):

Issue #5938, opened 2022-09-07
Sidebar fullscreen folder navigation


rmroot:
Sidebar should collapse when navigation is clicked to folder (this does work for assessments and calculators)

PR #5943, opened 2022-09-08
Issue 5938: Sidebar fullscreen folder navigation

rhernandez-intertech:
connects #5938


nbintertech (changes requested):
Try to use navigateSidebarLink() instead of collapseSidebar so the app can choose to not collapse if it's a desktop sized view. For home link and directory items, inventory items. For assessment items you can return the url+id string from goToAssessment.

If all the sidebar routing is done through navigateSidebarLink we can choose whether or not the sidebar needs to collapse (or when) in the future without having to change all the markup again.

nbintertech (approved):

Issue #5857, opened 2022-08-05
Apply mobile UI to Process Cooling calculators


rhernandez-intertech:

rhernandez-intertech:
I also configured the operating hours modal for mobile UI in this issue

PR #5950, opened 2022-09-12
Issue 5857: Apply mobile UI to Process Cooling calculators

rhernandez-intertech:
connects #5857


nbintertech (changes requested):
For small screens, we should reduce the size of the tab header text and/or add padding
image

nbintertech (approved):

Issue #5760, opened 2022-06-30
Cooling Tower Basin TH Integration


nbintertech:

rhernandez-intertech:
This calc uses Weather Data to generate a second set of results aside from the results generated from the calc itself, so we need to decided which set of results we want to use for Treasure Hunt, and if needed, restructure the calc to require the use of weather data

koay9f:
Can we hide the option to see the "See Calculated results"?
It is just kind of confusing
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/c92d9096-429e-4f13-ae8d-ae525ce4967b)

Issue #5759, opened 2022-06-30
Cooling Tower Fan TH Integration


nbintertech:

koay9f:
Getting an error from this one.

Also , the annual savings in the card do not match what is in the gauges
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/f69e596d-3827-44ec-a3c5-3c479e4af285)

koay9f:
Same secondary bug as energy perf & temp reset (lost gauges, haven't frozen yet...)

Issue #5758, opened 2022-06-30
Chiller Performance TH Integration


nbintertech:

koay9f:
Savings on card doesn't match gauges
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/d527813e-dbcc-4ee1-8583-73ed9e4fdf9e)

koay9f:
After clicking in the Edit Calculations then hitting "X" gauges disappear
and if you do it too many times the app seems to freeze (clicking doesn't work).
no errors
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/d04b8657-b08a-45cc-babc-ffde26727e37)

koay9f:
Just going to put this in here.  not sure which calc is the problem, but PPT export isn't working
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/2924b7a8-0150-4fe1-bb1b-62aee5edc2e5)

PR #5957, opened 2022-09-21
Issue 5760: Cooling Tower Basin TH Integration

rhernandez-intertech:
connects #5760 #5758 #5759


nbintertech (changes requested):
The issue with dissapearing gauges is still present. It may be as simple as forcing a window resize here (if you resize the window manually they appear). Check with Mark and see if he's run into any refresh/resize issues with those gauges before. Maybe something that's already been solved. If you don't find anything reach out to me **early** and I'll check it out too.

While you're in the game, please add padding between the weather data header tab and the TH text for save/cancel. Running hard into each other

nbintertech (approved):

Issue #5885, opened 2022-08-16
Low Priority Calculators for Training Release


nbintertech:
Last batch of calculators to apply responsive styles to

In order of priority:
1. Charge Mat
2. Cooling Loss
3. Opening Loss
4. Fixture
5. Atmosphere
6. Leakage
7. PH New calcs (waste heat, heat cascading etc);

koay9f:
This is looking a little weird
Charge Mat, Cooling, Wall, Opening, Fixture, Atmo, Leak

Low priority though.  It can go into the release like this if needed, but if can fix it before, awesome

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/07a27d19-ea6d-40b6-8495-937b48e37a20)

PR #5969, opened 2022-09-26
Issue 5885: Low Priority Calculators for Training Release

rhernandez-intertech:
connects #5885


nbintertech (changes requested):
Like the previous styling PR's we shouldn't duplicate html unless necessary. The best thing to do here would be to adjust styles so the original button group is flex-wrapped when out of space. the second best alternative would be to use a media query to apply vertical styles instead of duplicating html

nbintertech (approved):

Issue #5961, opened 2022-09-22
Steam 3 header bug


koay9f:
Problems
Med pressure causing problems. not saving.
Example seems to be okay...

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/4e42c4f0-66bc-4428-9838-fe9ac4ddd202)

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/dc8f50bb-6451-4e30-9400-666af64982a2)

If nav away and come back:

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/c7e5c88d-49e6-477e-98d4-96be0725a940)

Boiler
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/d7fd42d1-5807-4710-9023-cd87d913cc3a)
Ops - defaults. 1000 kW import

Will run with 2 headers

koay9f:
one validation went missing.  
Not absolutly necessary for release, but would be good (we have a steam training coming up)

Issue #5960, opened 2022-09-22
False warning in Header


koay9f:
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/89805786-c0e1-4c3a-aa39-3bbf308509a3)
Will go away if you navigate away and come back.
Not sure if is a problem in non-metric.
If you need the json let me know (just used data from the steam metric bug issue)

koay9f:
also - copy the steam example and change
`Flash Condensate Return` to "yes"
will get warnings (that also go away with navigation)

PR #5971, opened 2022-09-27
Issue 5960 5961

nbintertech:
#5960 #5961

-Fix multi header form validation
-don't validate when no constraint fields
-show validation message on form init
-capitalize Field names in validation like the rest of the app


src/app/ssmt/header/header.service.ts
```diff
@@ -98,8 +113,18 @@ export class HeaderService {
       useBaselineProcessSteamUsage: [obj.useBaselineProcessSteamUsage]
     });
 
+    // only add validation if pressureMin/Max constraint fields have value
+    if (pressureMin) {
```
rmroot, line 117: these should probably be if `pressureMin != undefined` right now it won't pass if `pressureMin == 0`

rmroot (changes requested):

rmroot (approved):

Issue #5869, opened 2022-08-11
Apply Mobile UI to Treasure Hunt


rhernandez-intertech:

koay9f:
The calculator title & save warning take up a lot of room. 
Maybe hide the warning? And the title and just have {icon} {Save} {X}? if it is really really small?

![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/60de8ac8-1e37-4fd7-9170-3248ef833ecb)

The Opportunity sheet is unusable.  
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/fd52c19a-01e8-4e02-bfee-47f5437f8313)

Ironically, the Opportunity sheet in the Treasure Chest looks fine, it's just the one within the calc that is being weird.

Setup & Treasure looks good
Treasure chest looks great

Results - mostly good

Issues that should not stop us from releasing, but if necessary can put in seperate issues to look at later (all graphs in reports, users can just move to a computer if they really want to see :D):
The pie chart at the end of the Exe Sum is ... too small to be seen?
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/1c919722-1f08-4da7-a382-ee01da84e72f)

Other graphs are also weirdly cut off
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/473e3dec-abd0-4f6c-8ed7-453be4ee1876)

This one is just adorably tiny
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/0b6e4056-5878-4e4e-8a43-b630f5e6534d)

This one is great, don't know what is special about it...
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/3c59ab2b-8d7a-49bc-a405-fab99792d81b)

I can't get "Payback vs Effort" to display in one screenshot, but it is pretty hilarious also (no fix for it, but toggle the labels for it to look even crazier!). Maybe for it, can you turn off the graph legend if it gets small?

PR #6012, opened 2022-10-31
Issue 5869: Apply Mobile UI to Treasure Hunt

rhernandez-intertech:
connects #5869


nbintertech (changes requested):
Changing width loses certain icons. This happens on develop also. Fix or hide the cell/container --> Work with Kristina to decide what should be happening here

image (52)
image (51)
image (50)

nbintertech:
Switching from TH chest to Find Treasure (or other tab navigation) sometimes takes single digit seconds to 35 seconds to resolve view. Was able to reproduce this twice, but not sure if there is a specific action causing it. I don't see any of your changes that could reasonably cause dthis, but please do some testing and investigate. Create a new issue if necessary

image (53)

PR #6042, opened 2022-11-15
Issue 5869: Apply Mobile UI to Treasure Hunt

rhernandez-intertech:
connects #5869


nbintertech (approved):
There are other calculators that have 'calculator-container' that are not calculators you're changing for this PR. They may have a header/h3 that falls under your css rule. Not sure if that's what your trying to achieve. Rule shuld have higher specificity if not.

Approved, check above and merge if fine.

Issue #6018, opened 2022-11-02
TH - PPT Improvements 


AlexBotts:
- Add slide for "Best Practices" 
- Add slide for "Other Opportunities Not Evaluated" 
- Add slide with savings gauges 
- Add carbon savings 

- Reduce the number of repetitive slides where possible

koay9f:
Should not delay release:
Best Practices & Other Opportunities -
Go ahead and have a text box for the page. Have it be bulleted list with text 
"* Outline key best practices identifed in Treasure Hunt here"
and 
"* Outline any major opportunities identified, but not evaluated here"

Savings gauges were not in 

I like the carbon donut  - maybe cut pie and combine the table slide with the donut (like teams)

maybe add cost donut also? (utility could get messy, maybe add electricty & NG if easy, otherwise just skip - maybe put just those two donuts on "Energy Utiltiy Usage & Cost" slide, skip the pie :D)

rhernandez-intertech:
replacing gauges with donuts

PR #6082, opened 2022-12-07
Issue 6018

rhernandez-intertech:
connects #6018


nbintertech (changes requested):
Code is fine, but I'm not sure if changes were suggested below, chart in the bottom right doesn't look quite right (utility/co2 selected).

image

nbintertech (approved):

Issue #5800, opened 2022-07-21
Data Explorer DayType Average


koay9f:
If user enters hourly data (or once we have 15 min and 30 min averages, if they enter 15 or 30 min data)
Then the whole thing gets kind of screwy (everything is excluded) b/c of there only being one data point.

My idea:

User ALWAYS enter their data interval (probably in seconds)
Before "Run Analysis" on DayTypes, dropdown with the 4 options used in CA module (15 min, 30 min, 1 hr, 24 hr)
maybe some error catching to make sure daytype time interval is not less than the data time interval

Honestly, some of this might be solved by fixing that other issue (0 hour assigned to the wrong day), but I think this will still help with teh 15 & 30 min data

nbintertech:
Note for verification, We/I need to test importing an old DE export with hourly averages calculated. _might_ need an adapter to make that work with the new intervals

koay9f:
As far as I can tell, this works really well (admittedly not checking math). My only gripe is that in DE it starts at 0:00 and in CA it starts at 0:30.
If this is too big for this go, we can close this issue and start a new one and save this but I think (for all data intervals) in CA, the time should start at 0:00
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/8226c9ef-ff4a-4b89-b9e8-4ac1501cc140)

koay9f:
Really should go in 1.2.4

nbintertech:
Issue #1 - Done
Issue #2 - I thought these start/ends might be fuzzy so I made it very easy to change in DE. I've changed it back to 0:30-23:30 so it matches CA. I'd say a new issue for changing CA system profile table, graphs all to 0-24 (and flipping DE back to match) could start being worked on for this release once everything else is done.

nbintertech:
- Fix default hourly interval
- Change start/end interval offset to match CA

koay9f:
Doesn't look like either fix made it over to develop.
ISSUE 1: CA is still {0+time interval} to 24:00 while DE is 0 to {24-time interval}.  I understand that DE is easier to change so let's change it (even though it is how we want it) for now and we'll fix them both with the other issue you created.
ISSUE 2: Dropdown still reverts to "Hourly" when you navigate away from page
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/96b6c4cb-e710-42a3-a21f-dd6095297bf6)

PR #6087, opened 2022-12-09
Issue 5800 - Day Type Average Intervals

nbintertech:
#5800 

Add 15, 30 min, 1 hr, 24hr intervals to DE for intergration with CA


rmroot (approved):
Looks good. Not sure it's applicable but this doesn't have any issues working with previously created CA assessments or day type analysis's does it? If they import from a previous versions of DE analysis is it going to break anything?

Issue #6102, opened 2022-12-13
Electricity Reduction Help text tweak


koay9f:
Add to power factor field help
Do not use this calculator to model savings from power factor correction without reducing the average current simultaneously.

QUESTION:
add yellow validation - if mod PF =/= baseline PF and baseline Current <= mod Current
"Power factor should not be reduced without also reducing the average current"

koay9f:
My fault - error message should read:
"Power factor should not be increased without also reducing the average current"

PR #6168, opened 2023-01-31
Issue 6102: Electricity Reduction Help text tweak

rhernandez-intertech:
connects #6102

### DIFF
--- a/src/app/calculator/utilities/electricity-reduction/electricity-reduction.component.html
+++ b/src/app/calculator/utilities/electricity-reduction/electricity-reduction.component.html
@@ -84,7 +84,7 @@
           <div *ngFor="let data of modificationData; let index = index;">
             <app-electricity-reduction-form [index]="index" [settings]="settings" [data]="data" [isBaseline]="false"
               (emitCalculate)="updateModificationData($event, index)" (emitChangeField)="changeField($event)"
-              (emitRemoveEquipment)="removeModificationEquipment($event)" [selected]="!baselineSelected" [userSelectedHP]="baselineData[0].userSelectedHP">
+              (emitRemoveEquipment)="removeModificationEquipment($event)" [selected]="!baselineSelected" [userSelectedHP]="baselineData[0].userSelectedHP" [modificationExists]="modificationExists">
             </app-electricity-reduction-form>
           </div>
         </div>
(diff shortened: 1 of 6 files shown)


src/app/calculator/utilities/electricity-reduction/electricity-reduction.component.ts
```diff
@@ -166,6 +166,8 @@ export class ElectricityReductionComponent implements OnInit {
 
 
   getResults() {
+    this.electricityReductionService.baselineData = this.baselineData;
```
nbintertech, line 169: Can move these to existing update___ methods. I guess double check the that the service wasn't already being updated.

nbintertech (changes requested):
Check validation logic vs what's in the issue. I believe we should be seeing validation below 
no-yellow

Other suggestions:
- Don't think modificationExists needs to be passed. We already have "isBaseline". if we are !isBaseline, call the method to check the warning.
- return the warning string or undefined if no warning -  instead of a boolean
- Make above changes and it should allow you to remove isValid as well as the isBaseline check in html check.
Generally just follow the pattern of other checkWarning methods

nbintertech (changes requested):

src/app/calculator/utilities/electricity-reduction/electricity-reduction.component.html
```diff
@@ -52,7 +52,7 @@ <h3>Baseline</h3>
         <div *ngFor="let data of baselineData; let index = index;">
           <app-electricity-reduction-form [index]="index" [settings]="settings" [data]="data" [isBaseline]="true"
             (emitCalculate)="updateBaselineData($event, index)" (emitChangeField)="changeField($event)"
-            (emitRemoveEquipment)="removeBaselineEquipment($event)" [selected]="baselineSelected" [userSelectedHP]="baselineData[0].userSelectedHP">
+            (emitRemoveEquipment)="removeBaselineEquipment($event)" [selected]="baselineSelected" [userSelectedHP]="baselineData[0].userSelectedHP" [warningMessage]="warningMessage">
```
nbintertech, line 55: BL form is set to not display warningMessage so don't need to pass input here

nbintertech (changes requested):
This is better. Conditions of warning work. The message still doesn't really make sense to me since it appears even when mod PF isn't reduced. I assume you're collabing with Kristina and she'll seee that in NV

nbintertech (approved):

Issue #6131, opened 2023-01-23
Save Annotations


nbintertech:
Save annotations for each chart so they appear when the user comes back into DE to view each chart

- don't need to save annotations from the current axis when selected axis changes (annotation data should keep reset when x axis changes)
- if Y axis changes, reset annotations
- if a Y axis is added, save annotations

PR #6184, opened 2023-02-07
Issue 6131: Save Annotations 

rhernandez-intertech:
connects #6131


nbintertech (changes requested):
I haven't tested this yet, but I think there will be more work needed here. Something to think about: is this method resetXAxisRelatedData still called, and if so is it still performing the needed function? 

If for some reason it's not needed, it should be removed.

Issue #6169, opened 2023-02-01
Sidebar should be available on assessments and inventory


nbintertech:
- Started this on branch issue-6127
- Default in assessment/inventory view is collapsed
- Also look at collapse functionality on app start - based on what size screen user has

nbintertech:
Most of the changes here were to adapt styles to the original to account for the sidebar width. Other changes:
- removed extra sidebar in report-rollup
- widened reports and removed redundant header
- when home link is clicked, sidebar automatically opens
- when assessment/inventory is clicked, sidebar collapses
- removed horizontal scroll in sidebar directory view and truncated at 25 chars

tested all print reports and they match the existing from develop, but there were a few with cutoff graphs - not sure if we have issues for those fixes yet

PR #6194, opened 2023-02-13
Issue 6169 - Show sidebar in every view

nbintertech:
#6169 

Largely stylistic changes to allow sidebar to work in every view
- a little bit of footer jump when header height changes but this is smoothed out quite a bit in the production build


src/app/dashboard/dashboard.component.ts
```diff
@@ -65,6 +97,10 @@ export class DashboardComponent implements OnInit {
       this.moveItems = val;
     });
 
+    this.showPrintViewSub = this.printOptionsMenuService.showPrintView.subscribe(val => {
```
rmroot, line 100: needs type and unsubscribe 😆

rmroot (changes requested):
image

Looks like this screen still needs to be addressed.

nbintertech:
Updated - if you've got any ideas as to why TH welcome-screen would need different styling vs say fsat welcome-screen considering they both have the same layout, css, and are top level components at the router-outlet... please do let me know so I can sleep this weekend.

The borders are like that in dev. I backlogged 6202 for that and more welcome-screen tightening

rmroot (approved):

Issue #6212, opened 2023-03-16
Leak calculator (TH) not updating when adding new?


koay9f:
From Paul's March 2023 Lear TH
3.	When editing multiple leak calculations, not sure how many, but screen stops showing the calculation details, like they are not there.  We had initially setup the facility to have compressed air utility, because it makes the leak calculations so much easier.  Then we converted them to electric once when we collected the logged compressor data.  After seeing this error, we re-load MEASUR, they re-appeared.  So they weren't deleted, but it sure looked that way.

I'm guessing by calculation details he means the list of leaks?
I may need to work with Paul a bit more to figure out what was going on here.  hopefully the assessment json will have some more clues.

nbintertech:
Wasn't able to replicate this same scenario, but I did fix a bug relating to deleting items. They may have made deletions, encountered an error and have continued to see broken results. If we get a more specific repro steps or problem continues I can search again.

PR #6227, opened 2023-04-11
Issue 6212 - CA leak survey issues

nbintertech:
#6212 

- Update leak index before resetting inputs
- Do cursor: pointer on hover for better UX


rhernandez-intertech (changes requested):
results tables do not update when a leak is deleted

Issue #6225, opened 2023-04-10
Data Explorer Multiple series segments


nbintertech:
Time series segments 'data' should have as many array items as the 'All Datapoints' segment.
Looks like additional data series are being ignored when createTimeSeriesSegments is called after a data series is added.

Issue #6177, opened 2023-02-03
Assessment Operations Pages - footer next button


nbintertech:
Some of the opp pages don't seem to be enabling/disabling next button correctly. Log tool/data explorer, psat, fsat for starters. Check all assessment/inventory types

PR #6228, opened 2023-04-11
Issue 6131 - Save annotations in memory

nbintertech:
#6131 

- Keeping comments with context/reasons why
- This needs #6225 to work with time segments

### DIFF
--- a/src/app/log-tool/log-tool-models.ts
+++ b/src/app/log-tool/log-tool-models.ts
@@ -226,6 +226,7 @@ export interface AnnotationData {
     x: number | string,
     y: number | string,
     dataSeriesId?: string,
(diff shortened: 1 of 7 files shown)


rhernandez-intertech (changes requested):
Annotations are not being saved when switching from graph to graph. When Y axis changes annotations do not reset

rhernandez-intertech:
Next button disabled even after data is uploaded
image

nbintertech:
> Next button disabled even after data is uploaded 
The changes in this PR shouldn't have sideffects on that logic. I think what you saw it probably #6177

rhernandez-intertech (approved):

Issue #6232, opened 2023-04-19
Pump Inventory: Summary


nbintertech:
Already started scaffolding some of this from #3798

Potential filter fields:
department, pump type, motor rated power, and status

![image (9).png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/d6206073-1568-4722-9760-3430eb1d259f)

Issue #3798, opened 2020-06-12
Pump Inventory: Setup


rmroot:
Create main setup tab with sub pages:
- Plant setup
- Departments
- Pump Properties
- Catalog

koay9f:
Overall things look good.  There are some issues with units that I'm wanting to check up on but will need some minor tweaks

nbintertech:
@koa I moved above comment to 6299. Feel free to close

PR #6240, opened 2023-05-01
Issue 3798, 6232 Pump Inventory

nbintertech:
#3798 #6232


rmroot (commented):
- Clicking "Pump Inventory" card on the home screen opens the modal to motor inventory

Otherwise looks great!

PR #6261, opened 2023-05-24
waste water mobile UI implemented

rhernandez-intertech:
connects #5873


nbintertech (changes requested):
Only thing I noticed - opening WW and expanding sidebar, next button appears over bar (play with z-index):
image

nbintertech (approved):

Issue #5874, opened 2022-08-12
Apply Mobile UI to PHAST


rhernandez-intertech:

koay9f:
The Aux, Design & Metered tabs lost the ability to switch between entering data and viewing help/results.  Copy format of calculators.

PR #6269, opened 2023-06-05
Issue 5874: PHAST mobile UI

rhernandez-intertech:
connects #5874


nbintertech (changes requested):
We should clean this up if possible. Seems like an invisible scroll is scrolling the outer container and we're going below the side bar. Only noticed in setup
image

Similarly let's get rid of double scrolls if possible
image


investigate or open an issue for energy equivelancy errors : 
image

Issue #6266, opened 2023-05-26
Turn off tutorials if app is not running in production


nbintertech:
Turn off tutorials by default if not in production. 

![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/18325076-aedc-4ba6-8013-cfe2281b7c5a)

Very low priority but this issue could be grown larger to look for other things that could be turned off or state changed to make development routines and testing easier.

nbintertech:
Put this one on the back buner for awhile. When you pick it back up there are a couple things to change:

We don't need to check whether tutorials have been disabled
I'd move this logic to the core.componet and execute once all data have been initialized
pull the global settings from the settingsDb service, make changes, and then update them in the DB (you can access them by dir id - example is in core.component

Closed PR and added comments here

kaichen127:
Leaving off my work on branch "issue-6266" - not able to figure out a way to turn off the main tutorial without entering the landing-screen.component

PR #6291, opened 2023-06-13
Issue 6266: Turn off tutorials if app is not running in production

kaichen127:
Connects #6266 

(is this good practice? should I be editing the global settings directly? )


rhernandez-intertech (approved):

nbintertech:
Closing - review comments added to issue

PR #6292, opened 2023-06-13
Issue 5874

rhernandez-intertech:
connects #5874 
fixes to PHAST mobile UI System Setup pages Aux Equipment, Designed Energy Use and Metered Energy, and System Setup nav banner


nbintertech (changes requested):
Looks like you can still advance the next arrow to the assessment when setup is invalid.

nbintertech (approved):

PR #6314, opened 2023-06-15
Issue 6015: Apply Mobile UI to Report Rollup

rhernandez-intertech:
connects #6015

### DIFF
--- a/src/app/report-rollup/assessment-reports/assessment-reports.component.ts
+++ b/src/app/report-rollup/assessment-reports/assessment-reports.component.ts
@@ -129,43 +129,22 @@ export class AssessmentReportsComponent implements OnInit {
     this.selectedFsatCalcs = selectedCalcs.filter(item => { return item.type == 'fan' });
   }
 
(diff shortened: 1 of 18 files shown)


src/app/report-rollup/assessment-reports/assessment-reports.component.ts
```diff
@@ -124,4 +128,49 @@ export class AssessmentReportsComponent implements OnInit {
     this.selectedPhastCalcs = selectedCalcs.filter(item => { return item.type == 'furnace' });
     this.selectedFsatCalcs = selectedCalcs.filter(item => { return item.type == 'fan' });
   }
+
+  goToAssessment(assessment: Assessment, mainTab?: string, subTab?: string) {
```
nbintertech, line 132: tab isn't being used, remove the conditions and assignment and just set the segment to navigate.

nbintertech (changes requested):
Has same issue as a few PR'S ago where the scroll is on the wrong container (I think)

image

Alert container -> can we add a little padding, reduce font by a few sizes, maybe play with left align... should improve readability here
image

Issue #6294, opened 2023-06-13
GA endpoint 502 error


koay9f:
502 errors occuring when navigating to various endpoints in app. Sometimes takes 30 sec -60 secs to appear. Will verify that we are sending a valid URL/data to weather api and go from there. As far as I can tell, analytics are still generally being received by GA 
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/cddb0d3e-b4c0-45ed-bef1-bcc0dd8b2839)

After talking with Gina.. there may be a timeout or retry issue going on. I was able to reproduce this after a few tries where I sat on the page for a bit, and the 502 err occured. Server logs also showing measur.ornl.gov as referrer, not just desktop

nbintertech:
Server side error from the weather api server that we also use for the desktop to send Google analytics. Not presenting in weather zipcode lookup

PR #6317, opened 2023-06-19
Issue 6294 GA endpoint issue

nbintertech:
#6294 

- Fix async call order problem
- modify inventory url's to track consistently
- remove extra production check


nbintertech:
Closing - may end up actually sending this as a solo commit to gina

rmroot (approved):

Issue #6295, opened 2023-06-13
Cooling Tower Basin Calc Weather Header Styling


nbintertech:
Fix 'Weather Data' header tab to take full width of container (but 50% of header). Also create better divide between result tab and left side of calculator whitespace.
![image.png](https://images.zenhubusercontent.com/5e4547eef6a311c23c81ce81/fb658d49-64fa-45d7-a53a-a1d4addc1513)

koay9f:
Weather data looks good but now the calc title is acting weird
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/57f60efa-8dd2-4c1f-890b-bb3f6f15c527)

PR #6321, opened 2023-06-20
Issue 6295: Cooling Tower Basin Calc Weather Header Styling

kaichen127:
Connects #6295


rhernandez-intertech (changes requested):
When in mobile view/smaller screen make the tabs at the top stack, like one on top of the other
image

nbintertech (approved):
Good, after Rachels request

Issue #6215, opened 2023-03-16
Calculate values in Facility Basics


koay9f:
From Paul's March 2023 Lear TH
10.	Under facility basics, it asks for energy price, quantity, and cost.  Ideally, if you input the quantity and either of the others, it should be able to calculate the average price or total cost. 
koa - I would have swore I already had this as an issue… but I cannot find it

nbintertech:
@kaichen127 give this a shot once your other sprint issues are done. Let's do a 10 min call first to outline.

PR #6324, opened 2023-06-22
Issue 6215: Calculate values in Facility Basics

kaichen127:
Connects #6215 

Also changed the checkboxes - previously a checkbox would show up that wouldn't work if the corresponding resource was saved in the treasure chest
image
before:
image
after:
image

### DIFF
--- a/src/app/treasure-hunt/operation-costs/operation-costs.component.ts
+++ b/src/app/treasure-hunt/operation-costs/operation-costs.component.ts
@@ -449,11 +449,14 @@ export class OperationCostsComponent implements OnInit {
   }
 
   calculateElectricityUnitCosts(){
+    this.checkElectricityForm();
     let annualElectricityCosts: number = this.treasureHunt.currentEnergyUsage.electricityCosts;
     let annualElectricityUsage: number = this.treasureHunt.currentEnergyUsage.electricityUsage;
     if (!annualElectricityCosts || !annualElectricityUsage)
     {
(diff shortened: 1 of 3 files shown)


rhernandez-intertech (changes requested):
looks good, but you need to make so that warnings go away once the form is valid without the user having to click on the estimate link again

src/app/treasure-hunt/operation-costs/operation-costs.component.ts
```diff
@@ -425,4 +448,463 @@ export class OperationCostsComponent implements OnInit {
     this.save();
   }
 
+  calculateElectricityUnitCosts(){
```
nbintertech, line 451: Initial glance...  I think there are a few opportunities to make things more DRY and clean up a bit. Take a look at template driven forms/forms through out the app and come back to this to improve.

Things to think about:
- Don't need the scoped annual___ variables defined at the start of each method
- Instead of using numerous check() methods, we _should_ be able to pass in args from the template to a single method that sets each variable representing error state. 
- Instead of "the rest of the form", tell the user specifically which fields need to be valid. 
- Is it necessary to track the error state of each property of each utility type or is there a way to simplify this?
- Anyway to genericize calculate methods or generally pull out repetitive logic

nbintertech (changes requested):

rhernandez-intertech (approved):

Issue #6287, opened 2023-06-09
Compressed Air - System Profile workflow


koay9f:
Should these two be the same for all day types?
![image.png](https://images.zenhubusercontent.com/5cd48a2af8cffa5a19122d27/c8e40939-6203-481a-99c8-c123ee4b3e3c)
If so, can we redo the flow so they don't have to set it for all daytypes (see error message by data entry section)

If not and in general, we need to make it more obvious that you are using this screen for all of the day types.  Maybe "next" changes the daytype? A warning if you have entered any data for a day type?

rhernandez-intertech:
here is a mock up of how CA system profile will look with a link to a list a Day Types and a warning message that will go away when all the Day Types have data entered.
The list of Day Types will appear in a popup modal, the list of Day Types will indicate which Day Types still need data to be entered.
![image.png](https://images.zenhubusercontent.com/60b7bed2e39f152beecc60b4/e8008ba6-36e8-4b9f-b961-4c16c28c9ee9)

PR #6333, opened 2023-06-27
Issue 6287: Compressed Air - System Profile workflow

rhernandez-intertech:
connects #6287


src/app/compressed-air-assessment/system-profile/profile-setup-form/profile-setup-form.component.ts
```diff
@@ -28,19 +28,23 @@ export class ProfileSetupFormComponent implements OnInit {
...
+      this.compressedAirAssessment = val;
       this.setPressureMinAndMax(val.compressorInventoryItems);
       if (val && this.isFormChange == false) {
         this.form = this.systemProfileService.getProfileSetupFormFromObj(val.systemProfile.systemProfileSetup, this.dayTypes);
         this.enableDisableForm();
       } else {
         this.isFormChange = false;
-      }
+      }  
+      this.checkDayTypesForData();
```
nbintertech, line 47: Should move this up under the if block. I think this is currently called an extra time if the compressedAirAssessment change originates from this form save(); other than that everything's gravy

nbintertech (changes requested):

PR #6367, opened 2023-07-11
Issue 6262: Apply Mobile UI to Compressed Air

rhernandez-intertech:
connects #6262

### DIFF
--- a/src/app/compressed-air-assessment/compressed-air-banner/compressed-air-banner.component.html
+++ b/src/app/compressed-air-assessment/compressed-air-banner/compressed-air-banner.component.html
@@ -233,8 +233,8 @@
             </li>
         </ul>
         
-        <div class="alert alert-info d-lg-none p-3 m-2 small">
-            The device currently being used is not the optimal size to fill out and/or view this page. It is recommended to use a device of 992px or lager.
+        <div class="alert alert-info d-lg-none p-1 m-1 small">
+            This page is optimized to be viewed on a tablet-sized device or larger.
         </div>
     </div>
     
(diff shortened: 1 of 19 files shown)


src/app/compressed-air-assessment/compressed-air-banner/compressed-air-banner.component.html
```diff
@@ -100,6 +191,51 @@
...
+            <li class="page-item" [ngClass]="{'active': secondaryAssessmentTab == 'graphs', 'disabled': !selectedModification}" *ngIf="secondaryAssessmentTab == 'graphs'">
+                <a class="nav-link small" (click)="changeSecondaryAssessmentTab('graphs')">Profile Summary Graphs</a>
+            </li>
+            <li class="page-item" [ngClass]="{'disabled': secondaryAssessmentTab == 'graphs' || !selectedModification}">
+                <a class="nav-link small" (click)="continueAssessmentTab()" [ngClass]="{'disabled': secondaryAssessmentTab == 'graphs' || !selectedModification}">
+                    <span class="fa fa-chevron-right"></span></a>
+            </li>
+        </ul>
+        
+        <div class="alert alert-info d-lg-none p-3 m-2 small">
```
nbintertech, line 236: I like that we're messaging them. We can condense this to:
A - "This page is optimized to be viewed on a tablet-sized device or larger" 
OR B - Find the closest standardized device resolution. I think we use 1024) and do "This page is optimized to be viewed on a device with ___ x ____ resolution or larger"

Should also be a shared component or some other angular mechanism to not have to change 5 locations. This would also actually be a great message to have in some of our other strained mobile views. Intern to do vibes.

nbintertech (changes requested):
Looking sharp

nbintertech (approved):

Issue #6312, opened 2023-06-15
Explore workaround for browser export/data limit


nbintertech:
Meet with me first to outline logtool/data explorer usage
1. Read/understand source code pertaining to #5834
2. Research options for working around problem (google, stack overflow, etc)
    a. export to other format (zip)?
3. Notes on what you find.

Meet to discuss

nbintertech:
Just noting that we originally started this for 5834 which is no longer an issue. Same logic can be applied to dashboard export/import

nbintertech:
I have a folder of big CA assessments made up for testing this fYI

PR #6368, opened 2023-07-12
Issue 6312: Explore workaround for browser export/data limit

kaichen127:
Connects #6312 
Installed pako and implemented gzip

some issues: 
if the user manually renames a file it won't be seen as a .gz file (or a .json file if that's what the user chose)
when any directory is selected (directory.selected = true) then none of its subcontents will successfully be exported (in both .json or .gz)


rhernandez-intertech (changes requested):

src/app/log-tool/export-modal/export-modal.component.html
```diff
@@ -22,14 +22,19 @@ <h3>Export Data</h3>
...
-                    <label for="exportName"> Name your file:</label>
+                    <label for="exportName"> Name your file:&nbsp;</label>
                     <input type="text" class="form-control w-100" id="exportName" [(ngModel)]="exportName">
                 </div>
             </div>
             <div class="modal-footer justify-content-between">
                 <button class="btn btn-secondary" (click)="closeExportData()">Cancel</button>
-                <button class="btn btn-primary ml-2" [disabled]="!dataExists" (click)="buildExportJSON()">Export
-                    as .json</button>
+                <div class = "d-flex">
```
nbintertech, line 31: We should simplify things for the user.
A. They probably won't be familiar with gz's, or file compression. I think we're better off not displaying any message at all, OR something similar to other long-processing areas of the app...  "Exporting to file... this may take some time depending on the size of your data" Something like that.
B. Similarly let's just have one button that says 'Export'. Our users don't really need to understand whether it's gz/json, unpacking a gz, or really using the files anywhere else (especially since you enabled them to import the same gz)

image (12)

nbintertech (changes requested):
With all the selections here (done with select all) the resulting gz doesn't import correctly.
image

image

image

src/app/dashboard/import-export/import-modal/import-modal.component.ts
```diff
@@ -42,21 +43,41 @@ export class ImportModalComponent implements OnInit {
...
-          this.validFile = true;
-        } else {
-          let fr: FileReader = new FileReader();
-          fr.readAsText($event.target.files[0]);
-          fr.onloadend = (e) => {
+          let blob = new Blob([$event.target.files[0]], { type: 'application/octet-stream' });
+          let arrayBuffer;
+          let result;
+          let fileReader = new FileReader();
+          fileReader.onload = () => {
```
nbintertech, line 52: I'd advise taking a second look at this block of fileReader calls from ln 52 on, debugging, making sure that the file content is coming in the correct format/encoding as it should be. Give it a look, reach out if you don't find the issue.

src/app/log-tool/data-setup/import-data/import-data.component.ts
```diff
@@ -175,19 +177,82 @@ export class ImportDataComponent implements OnInit {
     this.dayTypeGraphService.resetData();
     this.invalidFileReferences = new Array();
     if (files[0]) {
-        let extensionPattern: string = '.(json|JSON)$';
+        let extensionPattern: string = '.(json|JSON|gz|GZ)$';
```
nbintertech, line 180: The log tool part of the implementation can be pulled out since we won't need it anymore

PR #6390, opened 2023-07-19
Issue 6263: Apply Mobile UI to Motor Inventory

rhernandez-intertech:
connects #6263


nbintertech (changes requested):
These 4 chevrons - the background turns white when you click to the start or end and it becomes disabled. Looks good otherwise  
image

nbintertech (approved):

Issue #6355, opened 2023-07-03
Arrow key - Process Heating Fuel Fired


kaichen127:
Change amount field changes with single arrow key strike

HEAT BALANCE: 
![image.png](https://images.zenhubusercontent.com/6477700057b381ed6656b41c/c5e8390c-3395-4fde-af84-3242a3a87d3d)
![image.png](https://images.zenhubusercontent.com/6477700057b381ed6656b41c/93695f56-a5c9-42ba-b225-e23cb5c6033b)
![image.png](https://images.zenhubusercontent.com/6477700057b381ed6656b41c/9c7ea798-5158-4f71-8906-c4ba0e5b841e)
![image.png](https://images.zenhubusercontent.com/6477700057b381ed6656b41c/d8f2cd60-3c81-45e8-bb8a-99f1e01fe444)
![image.png](https://images.zenhubusercontent.com/6477700057b381ed6656b41c/cbd57ac5-e3e1-46b9-addb-1b0aeb041c72)
![image.png](https://images.zenhubusercontent.com/6477700057b381ed6656b41c/4bea2472-9e58-4f98-a931-cc004e3321f2)
![image.png](https://images.zenhubusercontent.com/6477700057b381ed6656b41c/fc4e0929-7191-4012-b787-5b16e63aa7e7)
![image.png](https://images.zenhubusercontent.com/6477700057b381ed6656b41c/2e145533-c0a4-4563-aceb-95f6cb91d08c)
![image.png](https://images.zenhubusercontent.com/6477700057b381ed6656b41c/1620979d-64db-4f92-8186-72d604ed8a44)
![image.png](https://images.zenhubusercontent.com/6477700057b381ed6656b41c/d204ee07-4696-4dff-bddc-74efd0bda1fd)
![image.png](https://images.zenhubusercontent.com/6477700057b381ed6656b41c/e5346ab9-f979-49ae-a40b-59891fc46895)
![image.png](https://images.zenhubusercontent.com/6477700057b381ed6656b41c/6a19e180-a3e1-4445-b7e9-6157238c7288)

AUX EQUIPMENT:
![image.png](https://images.zenhubusercontent.com/6477700057b381ed6656b41c/ba3cb684-f3bf-44f6-a214-f1f160310d10)
![image.png](https://images.zenhubusercontent.com/6477700057b381ed6656b41c/6569be2a-ed9e-49b4-a655-5c96c447110c)

DESIGN ENERGY USE:
[...]

PR #6401, opened 2023-07-25
Issue 6355: Arrow key - Process Heating Fuel Fired

kaichen127:
Connects #6355 

(also changed all instances of capital SCF to lowercase scf - might have changed some variable names too in the typescript but I separated them via commits)

### DIFF
--- a/src/app/calculator/compressed-air/air-flow-conversion/air-flow-conversion-help/air-flow-conversion-help.component.html
+++ b/src/app/calculator/compressed-air/air-flow-conversion/air-flow-conversion-help/air-flow-conversion-help.component.html
@@ -3,7 +3,7 @@
         Actual to Standard Airflow Help
         <br>
         <small class="text-muted">
-            The calculator converts ACFM (Actual cubic feet per minute) to SCFM (Standard cubic feet per minute) and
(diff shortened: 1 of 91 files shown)


rhernandez-intertech (changes requested):
Looking good, just a few were missed/needs correction: 
- Charge Materials - solid - Heat of Reaction needs to have a step of 10
- Flue Gas - solid/liquid - Ash Temperature step should be 10
- Other - Loss Description needs step of 10
- Auxiliary Power - Voltage step needs step of 10
- Auxiliary Power - Power Factor step needs step of 0.1
- Metered Energy - Fuel - Heating Value HHV needs step of 100

rhernandez-intertech (changes requested):
Other loss Loss Description still needs a step of 10
image

Metered Energy Heating Value HHV needs step of 100
image

rhernandez-intertech (approved):

nbintertech (changes requested):
There needs to be something in this issue and PR describing the scope of the unit changes here. This is a pretty widesweeping change that I don't see any requirements on - other than the unit label noted by Kristina. 

We'd normally put something like this in it's own issue since it's not related to number input steps, but since this is pretty big don't worry about pulling it out.

nbintertech (approved):

Issue #6441, opened 2023-08-11
Treasure Chest sort & filter menu rework 


rhernandez-intertech:
Rework/redesign the treasure chest menu to make the import and export buttons more obvious, move the sort and filter dropdowns into a consolidated dropdown to make room for other buttons

PR #6449, opened 2023-08-18
issue 6441: Treasure Chest sort & filter menu rework

rhernandez-intertech:
connects #6441


nbintertech (changes requested):
Existing work is good, but discussed additional changes to differentiate filter tags from all other buttons

Issue #5919, opened 2022-08-24
Moisture in Combustion Air Fixes


koay9f:
* Remove from Stack Loss Calculator UI (set to 0.73%)
* Update help text (reference that it is a percent)
* Work out actual min/max (there is a max amount of water that can be in air...)
* Maybe remove the `/100` and % symbol and just make a fraction

nbintertech:
BY VOLUME
Desktop unit in help text as %
Sent to suite as  fraction
```
SUITE METHOD
flueGasLossesByVolume --> 
getProcessHeatProperties --> expects fraction
```

BY MASS
Desktop unit in help text as %
Sent to suite as  fraction
```
SUITE METHOD
SolidLiquidFlueGassMaterial -->
getHeatLoss -- > expects %
```


Other methods used in calculator
BY MASS
calcExcessAir --> 
```
SUITE METHOD
flueGasByMassCalculateExcessAir --> expects %
```

```
SUITE METHOD
calculateFlueGasO2 --> expects %
```

PR #6452, opened 2023-08-21
Issue 5919 - Update moisture in combustion stack loss calc

nbintertech:
#5919 

Remove user field for moisture in combustion air
Fix expected unit sent to suite
update example vals

### DIFF
--- a/src/app/calculator/steam/stack-loss/stack-loss-help/stack-loss-help.component.html
+++ b/src/app/calculator/steam/stack-loss/stack-loss-help/stack-loss-help.component.html
@@ -14,6 +14,10 @@
         stack gas.
         <hr>
       </small>
+      <small class="text-muted">
+        Moisture in combustion air is assumed to be .73% for Stack Loss and Boiler Combustion Efficiency calculations.
+        <hr>
+      </small>
     </h6>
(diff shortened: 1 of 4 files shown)


src/app/calculator/steam/stack-loss/stack-loss-help/stack-loss-help.component.html
```diff
@@ -228,7 +228,7 @@ <h6>
       Moisture in Air
       <br>
       <small class="text-muted">
-        The moisture of the air (mass water / mass dry air). Can be determined with a psychrometric chart.
+        The percent moisture of the air (mass water / mass dry air). Can be determined with a psychrometric chart.
```
rhernandez-intertech, line 231: Since the moisture in air combustion field has been removed from the UI of the Stack Loss calc, this help text will never be seen by the users, so it needs to be moved to a section of the help text that is shown no matter what field is selected.

rhernandez-intertech (changes requested):

nbintertech:
Updated with help text per kristina. Kept the other help text in case it ever comes back

rhernandez-intertech (approved):

Issue #6264, opened 2023-05-24
Apply Mobile UI to Pump Inventory


rhernandez-intertech:

koay9f:
I clicked on the [home] icon and it wouldn't navigate and gave error.  
Only noticed this for pump inventory in mobile (other things worked in mobile, pump inventory worked in normal screen)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMVlVQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0d52725799a3cf82089195e06f7c49a242ce53cc/image.png)

PR #6514, opened 2023-10-11
Issue 6264

rhernandez-intertech:
connects #6264


nbintertech (changes requested):
Have pump inventory use the navigateHome()-->navigateWithSidebarOptions() like the other modules

nbintertech (approved):

Issue #6335, opened 2023-06-27
Mobile UI Explore Opps


koay9f:
In Expert view you haave the BL/Mod/Details Menu, but in Explore Opps you have to scroll to get to the results.
Would like to explore what a top menu for explore opps looks like

rhernandez-intertech:
also make the Help page a tab in the System Setup/Basics so users won't have to scroll down to the help text

koay9f:
Gauge in savings results is acting weird in pumps & fans & WW & CA. 
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMlFVQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a4b63fa76ad74fb41b1576ef3241c1740aba2abd/image.png)

Either shrink or just don't display.  Maybe ask in Thursday call? I vote don't display 
Process heating doesn't display it
Steam shrinks it

PR #6520, opened 2023-10-13
Issue 6335: Mobile UI Explore Opps

rhernandez-intertech:
connects #6335 
also removed the percent gauges in the results tables in PSAT, FSAT, SSMT, waste water and compressed air

### DIFF
--- a/src/app/compressed-air-assessment/explore-opportunities/explore-opportunities-results/explore-opportunities-results.component.html
+++ b/src/app/compressed-air-assessment/explore-opportunities/explore-opportunities-results/explore-opportunities-results.component.html
@@ -16,7 +16,15 @@
         <div class="d-flex align-items-center pl-1 pr-1 col-4">
             <span class="table-text"> &mdash; &mdash;</span>
         </div>
-        <div class="col-4">
+        <div class="col-4 d-lg-none d-flex align-items-center pl-1 pr-1">
+            <span *ngIf="(dayTypeModificationResult.allSavingsResults.savings.percentSavings) && (dayTypeModificationResult.allSavingsResults.savings.percentSavings >= 0)" class="table-text">
+                {{dayTypeModificationResult.allSavingsResults.savings.percentSavings | number:'1.0-1'}}%
+            </span>
+            <span *ngIf="(dayTypeModificationResult.allSavingsResults.savings.percentSavings) && (0 > dayTypeModificationResult.allSavingsResults.savings.percentSavings)" class="table-text">
+                0%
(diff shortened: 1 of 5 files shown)


src/app/compressed-air-assessment/explore-opportunities/explore-opportunities-results/explore-opportunities-results.component.html
```diff
@@ -16,7 +16,15 @@
         <div class="d-flex align-items-center pl-1 pr-1 col-4">
             <span class="table-text"> &mdash; &mdash;</span>
         </div>
-        <div class="col-4">
+        <div class="col-4 d-lg-none d-flex align-items-center pl-1 pr-1">
+            <span *ngIf="(dayTypeModificationResult.allSavingsResults.savings.percentSavings) && (dayTypeModificationResult.allSavingsResults.savings.percentSavings >= 0)" class="table-text">
```
nbintertech, line 20: For all the files with similar template block:
- There shouldn't be a need to group boolean operators if they're all AND / &&. Feel free to explain if there's something going on here.
- Doing (0 > whatever-val) is fine logically but 99% of the time I would express it as (whatever-val < 0)

src/app/psat/explore-opportunities/explore-opportunities-results/explore-opportunities-results.component.html
```diff
@@ -16,8 +16,24 @@
...
       <span class="table-text"> &mdash; &mdash;</span>
+    </div> 
+    <div class="col-4 d-lg-none d-flex align-items-center pl-1 pr-1" *ngIf="!inSetup">
+      <span *ngIf="(percentSavings && psat.modifications[exploreModIndex].psat.valid.isValid) && (percentSavings >= 0)" class="table-text">
+        {{percentSavings | number:'1.0-1'}}%
+      </span>
+      <span *ngIf="(percentSavings && psat.modifications[exploreModIndex].psat.valid.isValid) && (0 > percentSavings)" class="table-text">
+        0%
+      </span>
+      <div class="alert-danger small" *ngIf="!psat.modifications[exploreModIndex].psat.valid.isValid">
```
nbintertech, line 27: In the future,  it would be better to pass the opportunityPsatValid, selectedModification object, etc.. down into this component (or set it in this component), and check object exists, rather than indexing into the modification valid object. Nicer to read and one less point of failure to debug.

nbintertech (changes requested):

nbintertech (approved):

Issue #6478, opened 2023-09-20
Remove validation from Insulation calc's efficiency field


koay9f:
For both pipe and tank insulation... 
If possible, this only applies if the system is "Chilled". 
No upper limit on efficiency
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdE85IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--91cccc644857af314bd1a563ff0119d6430e058c/image.png)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdFc5IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--e812c7c28c7e44bf907720778b02cc1bd2446d85/image.png)

PR #6522, opened 2023-10-17
Issue 6478: Remove validation from Insulation calc's efficiency field

rhernandez-intertech:
connects #6478


src/app/calculator/steam/pipe-insulation-reduction/pipe-insulation-reduction.service.ts
```diff
@@ -72,6 +72,13 @@ export class PipeInsulationReductionService {
       heatedOrChilled: [{ value: obj.heatedOrChilled, disabled: !isBaseline }]
     });
 
+    if (obj.heatedOrChilled == 0) {
```
nbintertech, line 75: For both calculators - you can create a method on this service that adds validators to the form and returns it, based on val or heat or chilled. Use the service method in the form component and remove the logic there.

nbintertech (changes requested):

nbintertech (approved):

Issue #5538, opened 2022-04-18
Add Boiler blowdown rate calc to TH


koay9f:
Not sure they need much tweaking. Let me know if you need some help making them TH ready.
They would be "water" calculators (appear when water is selected, water filter, results are all water)

•	Boiler Blowdown Rate 
  (nbintertech - removed cooling tower since it's part of a new epic)

koay9f:
Can we actually hide this for now.
There is something wrong in the math (not unique to TH - the TH integration is perfect, but I don't want it in the TH while it is broken)

koay9f:
small tweaks
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMGpWQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b1dcbffd69aa1faaa99fc47918e0fbd2d33378fd/image.png)

koay9f:
Rachel - sorry for asking you to hide and now I'm going to ask you to unhide (please forgive me)
I think once hte small tweaks above are implemented (in the standalone & TH), this is probably good to go.

koay9f:
"Calculate Efficiency" - hitting "apply" does not fill in the boiler Effieiency field

Small tweaks "/hr" and change to example are good

PR #6550, opened 2023-10-27
Issue 5538: Add Boiler blowdown rate calc to TH

rhernandez-intertech:
connects #5538

### DIFF
--- a/src/app/treasure-hunt/treasure-hunt-calculator-services/boiler-blowdown-rate-treasure-hunt.service.ts
+++ b/src/app/treasure-hunt/treasure-hunt-calculator-services/boiler-blowdown-rate-treasure-hunt.service.ts
@@ -154,9 +154,9 @@ export class BoilerBlowdownRateTreasureHuntService {
         )
         utilityTypes.push(boilerBlowdownRate.baseline.boilerUtilityType);
 
-        let waterUnit: string = 'L/yr';
+        let waterUnit: string = 'gal/yr';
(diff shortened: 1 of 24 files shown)


src/app/treasure-hunt/treasure-hunt-calculator-services/boiler-blowdown-rate-treasure-hunt.service.ts
```diff
@@ -0,0 +1,200 @@
...
+                percent: ((baselineResults.boilerFuelCost - modificationResults.boilerFuelCost) / currentCosts) * 100,
+                label: boilerBlowdownRate.baseline.boilerUtilityType,
+                baselineCost: baselineResults.boilerFuelCost,
+                modificationCost: modificationResults.boilerFuelCost
+            }
+        )
+        utilityTypes.push(boilerBlowdownRate.baseline.boilerUtilityType);
+
+        let waterUnit: string = 'L/yr';
+        if (settings.unitsOfMeasure == 'Metric') {
```
nbintertech, line 158: I thin L/yr would be the metric value. If so, this should be reversed. More importantly, I don't really see where the 'energyUnit' value is being used in TH. It appears units are set at the display level, not at TH service level. So it's possible this code block is a meme that was once in use, or for a sub set of calculators but no longer is. Let me know what you find

nbintertech (changes requested):

nbintertech (approved):

Issue #6453, opened 2023-08-21
Add Assessment opportunity to Treasure Chest


nbintertech:
Add assessment opportunity (behave like existing custom opportunity):
- pumps, fans, CA, process heating 
- plug bl/mod energy use into util consumption
- allow user to select modification for opportunity 

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBZzloIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--e57981e4ece389c168dd179fda1b53a87dd368a1/image%20%2835%29.png)

koay9f:
If needed - can "turn off" process heating for the release. If you do shoot for getting PH included, then there are a few things than can be fixed later

Bugs found:
* Added Pump example to TH Example with the "VFD" scenario.  Turned off all opportunities except the Pump
     *  Cost was good. Utility Savings was good. CO2 savings was off. Was 229, should be 115. 
     *  Actually, if you turn off everything it is 115, so the opportunity might be okay, but something else is wrong (this was seen with other assessments also)
     *  Payback looks correct (it's not in other places, so was just checking here)
*  Added PH example to TH Example.  Tried to select "All Opportunities" scenario. Got error
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeWZHQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c08bdbf8fa2599a56b2d6e6600734c99b86d0f95/image.png)

     *   Added 2 zeros to NG use and cost so gauges were not mad
 *   Reset application, opened TH example, turned off everything, no savings
 *   Added pump example , selected vfd opp - CO2 savings still 229. Turned off the opp (nothing on) savings = 115t
 *   Reset again
 *   Increased the TH NG use & cost (x100), added PH assessment. no bug - seems to be related to not having enough NG in system setup.
      *   Does not have weird residual savings like pump
      *   Cost savings is correct (~$2M)
      *   CO2 savings is not (reading 3 t instead of 21,274.49)
      *   Payback is wrong (put total implementation cost of $7k).  (should be $7000 / $2004788/yr = 0.00349yr (probably should have made that higher costed :D )
[...]

PR #6559, opened 2023-11-03
Issue 6453 - Add Assessment Opportunities to TH

nbintertech:
#6453 

slack if you want assessments for test.

- adds modification id's to psat, fsat, phast
- assessment opps are very similar to standalone opportunity sheets. Tried to combine logic for these where it made sense ("custom opportunities"), and just duplicated code blocks with slight changes when necessary.
- bunch of small stray changes in unrelated files from bugs needing to be fixed to nav between assessment/th
- made issues to finish other assessment modules integration, as well as add modificationIds

### DIFF
--- a/scripts/download-stats.js
+++ b/scripts/download-stats.js
@@ -30,7 +30,6 @@ async function main() {
     }
 
     const headers = getHeaders();
-    debugger;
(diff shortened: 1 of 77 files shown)


rmroot (approved):
Integration Help is incomplete. Not sure if there is another issue for help or not.

Issue #6511, opened 2023-10-09
Percent Load Estimation


nbintertech:
Ensure chart updates when results update. See email/video submitted to Gina by user

gaccawi:
https://github.com/ORNL-AMO/AMO-Tools-Desktop/assets/8973004/20f731cd-32c5-410e-a3f6-d9be79ebec9a

nbintertech:
#6576

PR #6577, opened 2023-11-08
Issue 6511 - Percent load estimation - billboard.js bug 

nbintertech:
#6511 

fix chart breaking when results over 100%
delete redundant graph component
use same results for table and graph


nbintertech:
Closing - we DO need to see over 100%

Issue #6574, opened 2023-11-08
Pump Integration error


nbintertech:
Looks like something got broke 

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMHEyQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--273d90fa275fcfa45bda599911745426d89c3a32/image.png)

nbintertech:
branch 6572

PR #6578, opened 2023-11-08
Issue 6574 - Fix pump inventory integration in modify conditions

nbintertech:
#6574


nbintertech:
Not gonna bother with this yet since 6572 is going to require more

Issue #6509, opened 2023-10-06
Add "implementation cost" to "Explore Opportunities" Page


koay9f:
At the bottom or very top outside of the expand/collapse area

koay9f:
[implementationCostExpansion.pptx](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBemZBQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f96fddca46a37a08dfa71688cda87c660ffd26ac/implementationCostExpansion.pptx)

Issue #5460, opened 2022-03-16
Investigate native/template input validation on reactive forms


nbintertech:
Look into places we can safely delete validation attrs from the list below. If an Angular **reactive form** is being used (vs a template driven form) remove each attribute:

'min='
'max='

This fixes the app defaulting to native input validation in some situations. Behavior seems to have come after a recent update - possibly triggered by ngChangedAfterChecked or other timing issues

nbintertech:
This hasn't popped up anymore. Moving to icebox as a known issue

nbintertech:
Hasn't presented as an issue anymore

PR #6598, opened 2023-11-14
Issue 6509: Add "implementation cost" to "Explore Opportunities" Page

rhernandez-intertech:
connects #6509


nbintertech (changes requested):
Looks good. 

We need to avoid using native validation (min="__") on new implementations where we already have reactive form validation. It's caused problems in the past. Please remove any you see. #5460 

I see that implementation cost affects FSAT results where fields in other assessments do not. Unclear if adding payback as a result, for those that don't have it is supposed to be part of this issue.

rmroot (approved):

nbintertech (approved):

Issue #6551, opened 2023-10-31
Water/Wastewater units in Treasure Hunt


rhernandez-intertech:
Go through Treasure Hunt and make sure Water/Wastewater units are correct and consistent

rhernandez-intertech:
Will probably need @koay9f to help with this issue

PR #6639, opened 2023-11-28
Issue 6551: Water/Wastewater units in Treasure Hunt

rhernandez-intertech:
connects #6551
Steam units in TH were also made consistent


nbintertech (changes requested):
Found another 'kGal' at cooling-tower-treasurehunt-makeupservice

Also should just use the convert units method from kgal to gal

Issue #6696, opened 2024-02-22
Add Email assessment link to assessment banners


nbintertech:
Follow treasure-hunt-banner as an example, adapt shared modal if necessary

<img width="1332" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL3FCQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--340aebb9f10226610fd816ebc12211fd16abe1e0/image.png" alt="image.png" />

nbintertech:
- Added to all assessments and inventories
- added spinner for long requests
- Hidden in data explorer for now - The data is often over the reasonable size limit for server, or is at minimum pretty rough UX on a long request. Even after deleting uneccessary visualizer data that could be regenerated on import

nbintertech:
This got auto-closed because I used a github generated branch to connect tickets. Going to avoid doing that until setting are changed!

PR #6708, opened 2024-03-08
Issue 6696 - Add email attachment functionality to assessment and inventories

nbintertech:
#6696 

Also removing very old commented out code
data explorer email link intentionally commented out

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -688,17 +688,6 @@
         "node": ">=12"
       }
     },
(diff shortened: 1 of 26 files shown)


rhernandez-intertech (changes requested):
Since the emails are always going to be from the same sender email address, the 'From' field should be auto-filled with the email address and the field should be disabled or it doesn't need to be an input text field, just display the sender email address, because the way it currently is set up could be confusing to users 
image

rhernandez-intertech (approved):

Issue #6649, opened 2023-12-07
Steam - preheat makeup water with blowdown bug


koay9f:
When trying to make a scenario using the preheat makeup water with blowdown, it throws a bunch of errors - some steam property somewhere is failing the region select.

<img width="1597" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd3RCQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--101d3e0f3f764fa792c5aeff97abac140997ae40/image.png" alt="image.png" />

Weird thing is - take the file and change the BASELINE to have makeup preheat and it works...
<img width="1575" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdzFCQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--aacf4c4bceab531317a56faf8f62429bb79e769c/image.png" alt="image.png" />

[Steam Preheat Makeup with Blowdown Bug.txt](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdzVCQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--10b63c0c0699db96428bb45ebda879e1dc6b0248/Steam%20Preheat%20Makeup%20with%20Blowdown%20Bug.txt)

nbintertech:
It looks like #5369 was holding up but there were some other cases where the bleeding needed to be stopped, but we may want to go look at the validation errors in the suite that are causing errors to be returned, and decide whether MEASUR side validation should be adjusted

rmroot:
I have done some digging and the issue is coming from the "Preheat With Blowdown" modification in the test assessment. Specifically, it has something to do with the Condensing turbine. 

If you set the "Fixed Flow" of the condensing turbine to 16.1 klb then it works, 16.2 breaks the modeling. The results of the condensing turbine are used by the heat exchanger in this scenario.

Another note that makes the condensing turbine unique is the "Condenser Pressure" it uses a vacuum pressure unit settings and is converted to MPaa. that value is coming out very small.

koay9f:
[...]

PR #6733, opened 2024-04-09
Issue 6649 - Fix part of bad modeler output

nbintertech:
#6649 

Just adding some additional checks to stop app breaking on bad modeler output. Will still break report, but if Mark finds root cause of modeler error, and suite sends back appropriate object maybe we don't need to do anything else.


nbintertech:
Mark is working on root fix for this

Issue #6699, opened 2024-02-23
Fix: Users must hard refresh for new app release


nbintertech:
Users ideally shouldn't have to hard reset their browser upon new releases.

'Best' solution - use cache busting. In short, we'd add a release hash or some other identifier to the url so measur url would be treated as new. Need to study up on this . Any issues with current routing style? Additional server config needed from Hayden?

Minimal solution - Send user toast pop up on new release, asking them to refresh if they experience trouble

This will likely improve/fix some of the caching issues with #6706

nbintertech:
- Angular dist file and assets are setup to do cache-busting correctly (output-hashing)
- I've set up some event handlers to allow the app to tell when a new version has arrived. The suggested UX for this is popping up a toast notification with a button allowing the user to hard reload/refresh cache - rather than forcing reload.
- Won't be able to finish and test in a meaninful way until we have access to the dev server and I can download the PWA and test caching at a real URL

nbintertech:
Will close after 1.5.0 is deployed

koay9f:
I can actually see this in dev each time dev updates, but agree, will close after 1.5.0

nbintertech:
Working - tested going from 1.5.0 to 1.5.1

PR #6746, opened 2024-04-16
Issue 6699 - Send user toast message, allowing update PWA

nbintertech:
#6699 

Notify and allow user to update new MEASUR versions without hard refresh or reinstall


src/app/shared/pwa/pwa.module.ts
```diff
@@ -0,0 +1,12 @@
+import { NgModule } from '@angular/core';
```
rmroot, line 1: Module unused. Can delete.

rmroot (changes requested):
Looks good. Just need to remove unused module

rmroot (approved):

Issue #6721, opened 2024-04-03
Send analytics event on send email attachment feature


nbintertech:
When the send email attachment feature is used, send an analytics event so we can tell how often this feature is used.
- We should be able to tell which assessment module it was fired from. Use separate event names to track this, i.e. "event_TH, event_SSMT". there are many examples in app already.

First test locally using the 'weather' nodejs server.

PR #6762, opened 2024-04-22
Issue 6721: Send analytics event on send email attachment feature

rhernandez-intertech:
connects #6721


nbintertech (changes requested):
We still need to satisfy this from the issue --> 

> We should be able to tell which assessment module it was fired from. Use separate event names to track this, i.e. "event_TH, event_SSMT". there are many examples in app already.

nbintertech (approved):

Issue #6776, opened 2024-04-30
Tank Ins validation


koay9f:
<img width="907" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeU9sQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0baf3337262b559e854af078a2a762d27e85d257/image.png" alt="image.png" />

WHEN CHILLED
Surface Temp should be LESS than avg ambient Temp and greater than internal tank temp

koay9f:
Also - for the pipes one.  Pipe temp can be less than zero

PR #6784, opened 2024-05-07
Issue 6776: Tank Ins validation

rhernandez-intertech:
connects #6776


nbintertech (approved):
Users should see the validation message when an existing form is changed to chilled, in addition to the red input marker.
image

Issue #6759, opened 2024-04-18
Steam results - wrong data


koay9f:
The highlighted data points need to be x Operating Hours

Or the unit results needs to be "Energy/hr" instead of "Energy/yr", but since all the others are per year, I'd rather stick with that

<img width="855" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN0Z6QkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f10da0ade3f6566939c8bc71d006785cbe6d9bb7/image.png" alt="image.png" />

koay9f:
Forgot Stack Loss.  Everything else looks good
<img width="1627" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeXN4QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--66d5afb823534e45ea0cf4b626ed2e61462c5e62/image.png" alt="image.png" />

Issue #6758, opened 2024-04-18
Steam Results for Baseline


koay9f:
Unlike others, there is no results for steam BL shown.
At least in Turbines, if not headers there should be a results screen similar to that shown in Assessment, but with just the BL.

PR #6816, opened 2024-05-30
Issue 6758 & 6759: Steam results fixes 

rhernandez-intertech:
connects #6758 #6759


nbintertech (changes requested):
I'm not seeing any implementation/code issues here.

Should the sankey be changing? Work with Kristina to verify your results numbers and  check the sankey. This branch shows a very different version than dev. There is some unintented side effect going on in results:

image


in develop
image

nbintertech (approved):

Issue #6921, opened 2024-07-24
PF correction review


koay9f:
Notes from Chris:
Ok, first comment, can it be made so that you can change the name of the month? Something other than "Month 1", "Month 2", etc.?
"Annual Power Factor Penalty" in results should have a $

Tweak to accommodate new combination of data found at an INPLT by Chris
In the "PF Adjusted or Actual Demand" drop down, I'd add an option for something like "Both"
You'd have to add a third column here (koa - this sounds a little cramped to me, how does this look in mobile already?  I guess they don't need to be wide input boxes...)
 <img width="1725" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMEVJQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0b526b9ce4864d290e85a4113a7302bff53b439a/image.png" alt="image.png" />

The user would essentially enter this column as well, the rest of the results would be calculated normally (koa - in other words, do not calculate this column, but calculate the rest based on this data)
<img width="1725" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBME1JQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--afd47b6b2f4929c83752e44c189a23680e928bab/image.png" alt="image.png" />

(koa - he then doubts himself)
Does that make sense?

koay9f:
Re: editable fields -
Other option - new dropdown that is "First month" and has a list of the twelve months.  
Month 1 = "First Month"
Month 2 = next...
Add month --> whatever month is next

koay9f:
from chris: [2:56 PM] Price, Chris
Other note for the PF Correction calculator, need to add a warning for when you drop below 3 months of data
 
I would say you just shouldn't be able to drop below 3 months of data...

koay9f:
Chris requests that the "Delete Month" button be moved next to teh "Add Month Button" (since it is only deleting the last month anyway)
When you have "Both" can the order of the input table be: "Actual", "PF Adjusted Demand", "Power Factor" (so Power Factor is always on right)

koay9f:
Still there:
BUG:
FYI - things get buggy if:
how demand = "Real Power"
PF Adjusted/Actual Demand = "Both"
THEN
[...]

PR #6936, opened 2024-08-07
issue 6921: PF Correction calc fixes

rhernandez-intertech:
connects #6921

### DIFF
--- a/src/app/calculator/utilities/power-factor-correction/power-factor-correction-form/power-factor-correction-form.component.html
+++ b/src/app/calculator/utilities/power-factor-correction/power-factor-correction-form/power-factor-correction-form.component.html
@@ -111,7 +111,7 @@
     </th>
   </tr>
 
-  <tr *ngFor="let input of data.monthyInputs; let i = index;">
+  <tr *ngFor="let input of data.monthyInputs; let i = index; let last = last">
     <td>
       <input name="{{'month_'+i}}" type="text" step="any"
         [(ngModel)]="input.month" id="{{'month_'+i}}" (input)="calculate()" onfocus="this.select();">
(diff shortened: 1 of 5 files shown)


src/app/calculator/utilities/power-factor-correction/power-factor-correction-form/power-factor-correction-form.component.html
```diff
@@ -80,37 +94,48 @@
...
         (focus)="focusField('input2')">   
     </td>
-    <td class="btn-group btn-group-sm">
-      <button type="button" class="btn btn-danger pull-right btn-sm btn-form" (click)="btnDeleteMonth(i)">
+    <td *ngIf="data.billedForDemand == 0 && data.adjustedOrActual == 2">
+      <input name="{{'input3_'+i}}" type="number" step="any"
+        [(ngModel)]="input.input3" id="{{'input3_'+i}}" (input)="calculate()" onfocus="this.select();"
+        (focus)="focusField('input3')">   
+    </td>
+    <td class="btn-group btn-group-sm" *ngIf="i+1 == data.monthyInputs.length">
```
rmroot, line 134: Just a note. Not required change. I believe this is a check for if it's the last item in an ngFor. In an *ngFor you could do... 

`*ngFor="let input of data.monthyInputs; let i = index; let last = last;"` and then this would be `*ngIf="last"`

src/app/calculator/utilities/power-factor-correction/power-factor-correction.service.ts
```diff
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { PFMonthlyOutputs, PowerFactorCorrectionInputs, PowerFactorCorrectionOutputs } from './power-factor-correction.component';
+import { MonthyInputs, PFMonthlyOutputs, PowerFactorCorrectionInputs, PowerFactorCorrectionOutputs } from './power-factor-correction.component';
 
 @Injectable()
 export class PowerFactorCorrectionService {
```
rmroot, line 5: Another note that could be useful.. I would pull out all the "generate" functions that are just returning objects and move them into their own .ts file and just export that function and call it where it is needed. If you have any questions on that just let me know.

rmroot (approved):

Issue #6925, opened 2024-07-26
Automated Backup


nbintertech:
Push for upcoming release:
- Add sidebar link 'Data storage and backup'
- Add laymans Info about measur data storage and how it works #6924 , advise user to backup
- Allow restore from file
   - Desktop - Add automated backup file ion/off setting - file will be continously overwritten
   - Web - Adjust sidebar link and modal so user can download backup file of all data (without needing to select for export)
- Pop up warning if user navigates away without selecting automated backup location
- Pop up warning when user restores backup notifying them it will overwrite ALL data

Future:
- versioning?
- Refine UI during dashboard overhaul, possibly integrate to that area
- More direct control of the setting from sidebar

koay9f:
looks like a good start to me

Issue #6924, opened 2024-07-25
Add info about data storage to About page


koay9f:
Add section about where data is stored

Also, maybe add things about our open source license, etc?

Mark is doing the same thing for VERIFI, see what he's doing

https://dev.measur.ornl.gov/about
<img width="1479" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMWdPQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--fde3e95790230d075b5889e904ee44e276b6dd9c/image.png" alt="image.png" />

nbintertech:
Added info about data storage to 'data storage and backup' sidebar link. Added open source software declaration to about page

PR #6994, opened 2024-09-02
Issue 6925, 6924 - Add automatic backup feature

nbintertech:
#6925 #6924


rmroot (changes requested):
Overall the structure looks good. There's a few "todo" comments in here that you should probably double check if they need to be done before PR. 

The import backup modal component doesn't look finished

rmroot (approved):

Issue #7117, opened 2024-11-08
PF Triangle calc bug


rhernandez-intertech:
- something funky going on with the results where they equal the inputs
- set the chart to have a set x & y axis so the triangle change can be seen 
- show results below the chart

koay9f:
Using a static axis limit isn't really working well.
Can we do it so axis are something like 1.05 * max(realPower, reactivePower)
(sorry not sure of variable names)

Also, we need a quick pass at some validation
for now:
* power factor <=1
* power factor >0
* phase angle < 90

PR #7144, opened 2024-11-14
Issue 7117

rhernandez-intertech:
connects #7117

### DIFF
--- a/src/app/calculator/utilities/power-factor-triangle/power-factor-triangle-form/power-factor-triangle-form.component.ts
+++ b/src/app/calculator/utilities/power-factor-triangle/power-factor-triangle-form/power-factor-triangle-form.component.ts
@@ -36,7 +36,7 @@ export class PowerFactorTriangleFormComponent implements OnInit {
   }
 
   calculate() {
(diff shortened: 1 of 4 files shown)


src/app/calculator/utilities/power-factor-triangle/power-factor-triangle-form/power-factor-triangle-form.component.ts
```diff
@@ -36,7 +36,7 @@ export class PowerFactorTriangleFormComponent implements OnInit {
   }
 
   calculate() {
-    if (this.data.apparentPower != null && this.data.realPower != null && this.data.reactivePower != null && this.data.phaseAngle != null && this.data.powerFactor != null) {
+    if (this.data.apparentPower != null && this.data.realPower != null && this.data.reactivePower != null && (this.data.phaseAngle != null && this.data.phaseAngle <= 90) && (this.data.powerFactor != null && (this.data.powerFactor <= 1 && this.data.powerFactor > 0))) {
```
nbintertech, line 39: This could be quickly cleaned up. Can we ignore the null values and use convertNullInputValueForObjectConstructor() from the suite helper service if the suite is complaining? Otherwise set the not null chain and real validation logical statements as two variables.

nbintertech (changes requested):

nbintertech (approved):

Issue #7204, opened 2024-12-12
PF Correction into TH


koay9f:
Add PF Correction into TH
- Cost is in the form, need to add that into the Opportunity Sheet for them
- NO DIRECT ENERGY SAVINGS! just "Other cost" savings

If could have in before 1/19 that would be awesome

koay9f:
Looks great, but can we classify the opportunity as "Other"?  right now for implementation cost it is "Electricity" but savings it is "other" and that looks weird.

PR #7222, opened 2024-12-20
Issue 7204

rhernandez-intertech:
connects #7204


nbintertech:
FYI looks like when you open a saved opportunity, the opportunityCardData passed to setCalculatorInputFromOpportunity() is missing the inputData property and is breaking. Not sure if this is going to be done elsewhere due to the logical difference with this calculator

nbintertech (approved):

PR #7254, opened 2025-01-10
Issue 6806

rhernandez-intertech:
connects #6804 #6806 #6808


nbintertech (changes requested):
Need's errors to be fixed or silenced to be ready for a proper review. Can discuss more over a call

nbintertech (approved):

Issue #7270, opened 2025-01-16
Plant Setup form 


rhernandez-intertech:
create the setup form and help text

PR #7272, opened 2025-01-21
Plant setup form added to CA inventory

rhernandez-intertech:
connects #7270


nbintertech (approved):
2 problems below. I would comment or block these out until you've broken down the work to completely fill out the setting/unit conversion logic that all assessments have
'update data' after a units change seems to generally not work
'electricity cost' doesn't save

I'm not sure CA inventory will be 'three-way-connected' like pumps is (pump to motor to psat)

You might do best to ignore and comment out a lot of the connected inventory logic as much as is possible until all the assessment's own operations are fleshed out

Issue #7177, opened 2024-12-04
Update existing toasts 


nbintertech:
Update existing toast messages with style from the user survey feature. Create a styled dumb component that toast content can be dropped into. I believe it's currently just version update toasts and browser data notice toast

Let's also revamp language on these:
- specifically the 'install' when user is in web version (per Tom)

nbintertech:
I'll verify once pushed to dev server. UI changes are in PPT

Updated, cleaned and reorganized a number of files/logic code side, but main UI updates are:

- Application update toast notifications follow new style
- New messaging for web/pwa update notification
- Added lower right snackbar notification for quick info/error notifications
- Moved sidebar version links nearer to top
- Update release notes modal with cleaner measur-green header and text styling.

[UI - update workflow and notification touch ups.pptx](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMGVQQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--58b90683579fbad4ce303be6406a0b2b19e289e0/UI%20-%20update%20workflow%20and%20notification%20touch%20ups.pptx)

PR #7279, opened 2025-01-23
Issue 7177 - Update Modals with toast styles, add snackbar notification, sidebar links, reorganize, cleanup 

nbintertech:
#7177


rmroot (approved):
Looks good. The only thing I would say is to make sure that these are hidden on print. I tested the psat and fsat reports. They have a little line on some of the pages. Might be pre-existing or from these new items. Just verify that they aren't new

Issue #7404, opened 2025-03-20
End Uses refactor 


rhernandez-intertech:
Nest End Uses within Systems (each System has their own list of end uses)
move Leak Rate and Known Total Airflow to System Information 
On the End Use page add tabs for each System

PR #7428, opened 2025-04-02
Issue 7404

rhernandez-intertech:
connects #7404


src/app/compressed-air-inventory/compressed-air-inventory-setup/end-uses-table/end-uses-table.component.ts
```diff
@@ -23,29 +24,49 @@ export class EndUsesTableComponent implements OnInit {
   deleteSelectedId: string;
   confirmDeleteData: ConfirmDeleteData;
 
+  selectedSystemIdSub: Subscription;
+  selectedSystemId: string;
+  selectedSystemIndex: number;
```
nbintertech, line 29: I can't see a reason why we need to track selectedSystemIndex here. We already have the systemId from the .behaviorSubject. We can use that to get the system, check that it exists and perform operations. Indexing into an array and chaining another operation (particularly inside of an array property of that object is more error prone).

nbintertech (changes requested):

nbintertech (approved):

PR #7435, opened 2025-04-04
Issue 7400: End Uses pie chart

rhernandez-intertech:
connects #7400


nbintertech (approved):
Commit containing pie chart implementation is good to go. The other two commits have review from the previous PR

Issue #7407, opened 2025-03-20
Move pre-assessment *calculator* from general to PH


koay9f:
just move the calc, no action for the dashboard version.

PR #7439, opened 2025-04-07
Issue 7407: moved Pre Assessment to PH

rhernandez-intertech:
connects #7407


nbintertech (changes requested):
1. Calculator link is still listed in Sidebar > General, but not in Sidebar > furnace
2. If pre-assessment is furnace related (as it seems) move it to that module instead of importing from utilities.

nbintertech:
Point 2 on original comment - what's the story here?

Issue #7316, opened 2025-02-06
Cash Flow Calc upgrade


koay9f:
Let's put some math in!

1 - Rename "Annual Fuel Costs" to "Other Annual Costs"
2 - Add "Other Annual Savings"
3 - Add advanced cashflow data entry:
* 2 fields - year (max = lifetime), cashflow
* has an "add cashflow" button that adds another row

4 - lots of math (see attached), I can help walk through. All can be done in typescript unless we think the iteration list will be long or we just want the math to be in the suite
5 - new results 

[cash flow IRR.xlsx](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL3JDQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d332a9df66db514931703b5a289a5db930fedc9f/cash%20flow%20IRR.xlsx)

koay9f:
Results are generally kind of wonky

* not getting a value for IRR - even in example - if can't get to work in time, hide the result for the moment
* I think you need to have the "Other Cashflows" on for results to work - maybe for now we skip the "Add Cashflow" button and have them always there
* Other savings dos not seem to be hooked up to results, same with Other costs* 


HELP TEXT
Annual Energy Savings
Annual expected reduction in energy costs resulting from the energy efficiency measure.

Other Annual Savings
Annual expected reduction in other costs or additional revenues resulting from the energy efficiency measure


Installation Cost
The expected cost to complete the energy efficiency project.

Annual Operation Cost
Annual operation costs (including maintenance and other yearly costs) for the equipment or increase in operation costs with the new equipment (excluding fuel). (If there is an operation cost savings, that should be include in "Annual Other Savings").

Discount Rate
The return that could be earned per unit time on an investment with similar risk. It is used to convert present or annual cash flows to the other. 
It can reflect the opportunity cost of investment/savings, the riskiness of the project, and the time value of money. 
A higher discount rate reduces the present value of future cash flows, while a lower discount rate increases it.

PR #7457, opened 2025-04-18
Issue 7316: Cash Flow Calc upgrade

rhernandez-intertech:
connects #7316


nbintertech (changes requested):
Nice work on this. Calculations and logic are well structured. Two improvements to work on before merge:

1. Go through cash flow service and make a reasonable attempt to touch up readability. 
Ex. starting at line 222, the results being assigned here could be broken up into intermediate vars, where 2-3 readable vars/lines represent the assigned result instead of a horizontally long compound formula full of parens. Another example, rename "brute force" related logic and vars. "Brute Force" is just a type of algorithm, so what are we solving for here? Seems to me like annual cash flow by interest rate. Same notes for any variable names.. sum1, sum2, iteration_, etc.

3. Center and 50-75% width on advanced cash flow table

PR #7469, opened 2025-04-23
Issue 7316: Cash Flow Calc upgrade

rhernandez-intertech:
connects #7316

### DIFF
--- a/src/app/calculator/utilities/cash-flow/cash-flow.component.ts
+++ b/src/app/calculator/utilities/cash-flow/cash-flow.component.ts
@@ -3,7 +3,6 @@ import { IRRBruteForceResults, CashFlowFinalResults, CashFlowForm, CashFlowOutpu
 import { CashFlowService } from './cash-flow.service';
 import { SettingsDbService } from '../../../indexedDb/settings-db.service';
 import { AnalyticsService } from '../../../shared/analytics/analytics.service';
-import { c } from '@angular/core/navigation_types.d-u4EOrrdZ';
 
 @Component({
   selector: 'app-cash-flow',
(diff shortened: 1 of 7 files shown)


src/app/calculator/utilities/cash-flow/cash-flow.component.ts
```diff
@@ -3,6 +3,7 @@ import { IRRBruteForceResults, CashFlowFinalResults, CashFlowForm, CashFlowOutpu
 import { CashFlowService } from './cash-flow.service';
 import { SettingsDbService } from '../../../indexedDb/settings-db.service';
 import { AnalyticsService } from '../../../shared/analytics/analytics.service';
+import { c } from '@angular/core/navigation_types.d-u4EOrrdZ';
```
nbintertech, line 6: Remove. I assume these are coming from copilot autoimports.

src/app/calculator/utilities/cash-flow/cash-flow.service.ts
```diff
@@ -207,15 +212,19 @@ export class CashFlowService {
     let sumContinueB: number = 0;
     let sumIterationB: number = 0;
     bruteForceResults.forEach(result => {
-      sumContinueA += result.continueA;
+      //sumContinueA += result.continueA;
```
nbintertech, line 215: Block from 214 on: 
1.NaNs can likely be prevented earlier on so you don't need the utility method. I would guess they're being introduced where you're incrementing in this loop.
2.There is redundancy re: iteration/loops in ln 221 and on. If we only need the continuelists to sum them up, JS array methods like reduce() could be used to make things clearer.

nbintertech (approved):

nbintertech (changes requested):
remove the import

nbintertech (approved):

Issue #7438, opened 2025-04-07
[BUG] No name end uses unrecoverable


nbintertech:
**Describe the problem**
When end use name becomes empty, the user can click on a different end use. This puts the no-name end use in an unrecoverable state where it has to be deleted.
This will also need to be fixed in CA inventories
**Expected behavior**
If table contains any with no name, populate with temp name so it can be clicked

**To Reproduce**
Go to the example Compressed Air Assessment --> End Uses tab --> add an end use --> remove name and click a different end use

**Additional context**
<img width="1701" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNE9TQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--dd8815989c04e09921b9db97356a8e54c7074071/image.png" alt="image.png" />

PR #7489, opened 2025-04-30
Issue 7438: No name end uses unrecoverable

rhernandez-intertech:
connects #7438

### DIFF
--- a/src/app/compressed-air-assessment/results-panel/end-use-table/end-use-table.component.html
+++ b/src/app/compressed-air-assessment/results-panel/end-use-table/end-use-table.component.html
@@ -18,7 +18,7 @@
             <tbody>
                 <tr *ngFor="let endUse of compressedAirAssessment.endUseData.endUses"
                     [ngClass]="{'selected-end-use': endUse.endUseId == selectedEndUse.endUseId,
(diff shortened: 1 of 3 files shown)


src/app/compressed-air-assessment/results-panel/end-use-table/end-use-table.component.html
```diff
@@ -18,7 +18,7 @@
             <tbody>
                 <tr *ngFor="let endUse of compressedAirAssessment.endUseData.endUses"
                     [ngClass]="{'selected-end-use': endUse.endUseId == selectedEndUse.endUseId,
-                    'invalid-end-use': !endUse.isValid}">
+                    'invalid-end-use': !endUse.isValid}" (click)="selectItem(endUse)">
```
nbintertech, line 21: We can do better. The cursor value should be pointer on hover. The user needs something to click on. Display the default name if there isn't one. CA inventory needs it as well.

nbintertech (changes requested):

nbintertech:
Good. Compressor Inventory has the issue also

nbintertech (approved):

Issue #7531, opened 2025-05-20
Add copy table function to all tables in reports


koay9f:
Specific request: Compressed Air: Exe Summary & payback details
can turn this into multiple issues

PR #7586, opened 2025-06-18
Issue 7531: Add copy table function to all tables in reports

rhernandez-intertech:
connects #7531


nbintertech (changes requested):
Some of the success notification windows are popping up half hidden by footer, or overlapping other text (looking janky with their opacity). In the inventories, and CA -> baseline >System profile tables for example.

See if we can make a reasonable effort to fix that styling. If it's going to be a chore given the scope of this, make Kristina aware and we can create a backlog/intern issue

Issue #7678, opened 2025-08-05
Remove Process Heating SQL lite dependency


nbintertech:
TBD

nbintertech:
merged to `remove-sqlite`

PR #7701, opened 2025-08-13
Issue 7678 - Remove SQLlite Wall Loss

nbintertech:
#7678 
Adding more detail than usual here as we'll base future removals off this.

Priority:
- Remove all SQLDb references and calls for the given material
- Add material method for inserting default materials on startup
- Replace with calls to _WhateverMaterial_DbService in PHAST, and calculator, SuiteDb, Custom materisl

Tech debt upgrades - SuiteDb component (Wall Losses Surface):
- Add new method on customMaterialsService to setMaterialNameError, Remove method definition for check material name in component 
- Remove lodash in transformation/filters and replace with js native methods
- Use initMaterials() and set a default selected material
- remove canAdd
- remove isValidMaterialName
- Change name to ‘Custom’ over ‘mod’
- group updateMaterialsAndclose logic into method
- Overall reduce template complexity, remove Input + change name to original material

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4365,14 +4365,14 @@
       }
     },
     "node_modules/@inquirer/core": {
-      "version": "10.1.10",
-      "resolved": "https://registry.npmjs.org/@inquirer/core/-/core-10.1.10.tgz",
-      "integrity": "sha512-roDaKeY1PYY0aCqhRmXihrHjoSW2A00pV3Ke5fTpMCkzcGF64R8e0lw3dK+eLEHwS4vB5RnW1wuQmvzoRul8Mw==",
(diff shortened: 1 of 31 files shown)


src/app/indexedDb/wall-losses-surface-db.service.ts
```diff
@@ -15,6 +16,27 @@ export class WallLossesSurfaceDbService {
 
   }
 
+  insertDefaultMaterials(): Observable<number[]> {
```
rmroot, line 19: Should DefaultData.delete() be called?

any reason we are not also bringing in the id in the defaultMaterials?

Issue #7661, opened 2025-07-17
Steam - Boiler - Temperature input functionality upgrade


koay9f:
before Steam Temp have dropdown/toggle for Saturated vs. Superheated.  Default to Saturated.

if saturated
ask if they want to supply temperature or pressure (bonus points if you have the "field name" be a dropdown itself instead of it operating like the saturated calc does now - and if you go for bonus points, more bonus points if you make an issue to do the same in the saturated and superheated steam calcs)
show the resulting pressure / temperature using saturated steam calculator (bold result)
on HEADER page, have the pressure for the highest header be a result instead of an input and have it be the user input pressure / calculated pressure from the boiler page

if superheated
ask for temperature and pressure
on HEADER page, have the pressure for the highest header be a result instead of an input and have it be the user input pressure   from the boiler page

WoodrowKromar:
Boiled down Criteria:

**Boiler page**

1. Added Steam Quality dropdown with selections Saturated, Superheated. IF Saturated  Display Saturated Pressure, Saturated Temperature dropdown in-line with accompanying temperature/pressure input field that is toggled on the dropdown selection. Only one may be showing. IF Superheated, display both temperature/pressure input fields. 
2. Display Pressure and Temperature as additional readonly fields, that return the calculated values.

**Header Page**
1. Display the readonly calculated values of Pressure and Temperature that are calculated on the Boiler page.
2. Only the High Pressure Header displays the above fields.

Bonus

**Saturated Properties Calculator**
1. In-line dropdown and input field of Saturated Pressure or Saturated Temperature.

**Steam Properties Calculator**
1. In-line dropdown and input field that is tied to the selection of the Input field. Previously known as Known Variable.

koay9f:
This all looks good to me. With Tom and Riyaz for more crazy testing. but closing this for now

nbintertech:
Needs a fix to show calculated outputs on switch to saturated.
This needs a final check by me,  end to end, for all scenarios, given every change we've made in the last few days.

PR #7898, opened 2025-10-15
Issue 7661

WoodrowKromar:
#7661 

Reordered the steam-properties and saturated-properties calculator so two fields take un one row. Fields were too long originally and don't need to take up all the space.


nbintertech:
@WoodrowKromar 

- Updated SSMT to use the event emitter already in place.
- Added Validation (this is waiting on direction, see issue comment)
- removed some unused identifiers/logic
- added quality of life mechanics (form controls getters, enum for the number controls). 

Feel free to continue from these changes

Issue #7895, opened 2025-10-15
[BUG] Inventory: Connected Items bug


nbintertech:
**Describe the problem**

Error on Disconnect motor when motor is integrated with Pump or CA inventory. Looks like a symptom of a wider issue.

**To Reproduce**
Connect a pump or ca inventory to a motor. Visit the connected motor and return to the inventory to disconnect the item

### Higher priority
- Error on Disconnect motor - ca-motor-int service 249 (this is also an issue with pumps. will break out if necessary)

<img width="1213" height="233" alt="Image" src="https://github.com/user-attachments/assets/427525ce-5f8e-458b-97cb-5e9ba648f127" />

PR #7911, opened 2025-10-21
Issue 7895 - Adjust conversions for bag method specific results, and other leak results in air leak survey

nbintertech:
#7895


WoodrowKromar (approved):
Skimmed and looks good. Approving

Issue #7907, opened 2025-10-20
Power Factor Calc Starting Year Applies to Entered Months


koay9f:
Thank you for adding the "Power Factor Correction" calculator.  A few notes:

•	Changing the "Starting Year" does not change the year in the tables
**Below is now part of issue #7922.**
•	For months where the power factor exceeds the mandated minimum power factor, the calculator results have negative values for "Demand Penalty" "Penalty Cost" and "Proposed Capacitance" and incorrect values for the "Proposed Reactive Power".   Additionally, the proposed fixed capacitance is given as a negative value, where it probably should be zero (with all capacitance being variable capacitance).
{going to ask Chris for help with this one

<img width="576" height="229" alt="Image" src="https://github.com/user-attachments/assets/71e316d6-d478-45b4-9968-dc03c6f54f9c" />

from Bryan Rasmussen  10/15/25

PR #7921, opened 2025-10-24
Adjustment of the Starting Year updates each monthly entry

WoodrowKromar:
#7907 Power Factor Correction.

Adjusting the Starting Year now updates the values in the editable table directly below, which calculates and updates the values in the readonly table on the right.


WoodrowKromar:
Original Issue will be split per Kirstina Directions and this solution will meet the current issue criteria.

nbintertech (changes requested):

nbintertech (changes requested):
Since there are other changes, you may also check that year is not null in setMonthNames()

Issue #7565, opened 2025-05-29
Pump Fixes


nbintertech:
Update app-wide validation for pump stages
Update pump inventory stages defaults to be 1

Let's walk through when you start

koay9f:
So, you're looking through out the app (pump module, pump inventory, pump calculators) for this field:

<img width="763" height="651" alt="Image" src="https://github.com/user-attachments/assets/2aec1e99-162b-4642-ae7c-cee136970fed" />

It's default value should be 1.  The field should like the field there (with the nice buttons).  
Validation - 
Red: it should be >=1.  it should be a whole number. 
Yellow: < 15 "Multistage pumps usually have less than 15 impellers. Please check your data."
I think that's it...

koay9f:
Honestly, I'm getting values below zero before this error appears so I think the validation isn't triggering as early as it "should" given that wording. It seems to be triggering when the efficiency for the points hits below zero.
Let's change text to "Efficiency at specified points is below zero, please check data inputs"
If this doesn't get done before release, that is fine - the work done so far can stay in, let's just finish up for next release.

PR #7977, opened 2025-11-20
Pump Stages Updated

WoodrowKromar:
#7565 
- Min value by "+" or "-" set as 1 instead of 0. Validation if num does reach 0 is still in place
- If Staging goes higher than 14, we yellow banner warn that 15 and higher is unusual.
- Values now must be a whole num.


nbintertech (changes requested):
1. Move the pattern validation to a shared folder custom validator for whole number validation. Use GreaterThanValidator as an example.
2. Remove setValidation() methods. Add the validation pattern/validator to the ValidatorFN array that's passed to the form control in formBuilder.group() options. Do this for each method that builds and returns the form object. Example for pump-equipment-catalog --> getFormFromPumpEquipmentProperties()
3. Create a constant for the warning value and store it in psatConstants.ts or another pumps ts to import at each location. 
4. We can remove the html native attribute validation for min
5. In Achievable Efficiency there is a side-effect due to the stages being to high. Seems to be anything higher than 9, but this might be contextually dependent on other fields. Address  this last.
image

If you have questions on any of the "why" related to above, let me know.

nbintertech (approved):

Issue #7086, opened 2024-10-24
Water: Small screen styles and touch ups


nbintertech:
1. Water Assessment - Implement small screen styles as is done for other assessment modules.

  Example.. styles needed for:
- Assessment setup and water system tabs
- Chevron arrow back/continue functionality
- Continue to allow navigation to the diagram tab when tablet size or smaller screens but block diagram rendering and include an info banner/message  "Best viewed on x screen sizes.." as we have done elsewhere. Search "Best viewed" should turn up examples

2. Water Diagram - Style Touch ups:
  - Select Known Loss component -> Edit drawer --> Styles --> component color does not appear to do anything and should be fixed to set color. Hide handle customization if it is not already hidden (Screen A). Ensure "delete component" is still available
  - Truncate input names when they are overflowing under the units (Screen B). This is just adding polish. Do not spend too much time on. If it becomes fiddly styling work, we can omit it from this issue.

Screen A
<img width="1304" height="575" alt="Image" src="https://github.com/user-attachments/assets/56f11cc2-7b82-4e85-8fb8-c9d8dfbc8a0f" />

Screen B

<img width="324" height="73" alt="Image" src="https://github.com/user-attachments/assets/59566d4e-b882-40a5-b76a-5344aa2cc02a" />

WoodrowKromar:
@nbintertech Regarding 2. Water Diagram - Style Touch ups:

Issue: Truncate input names when they are overflowing under the units (Screen B). This is just adding polish. Do not spend too much time on. If it becomes fiddly styling work, we can omit it from this issue.

I cannot replicate the overlapping truncation.

PR #7984, opened 2025-11-24
Style Updates and React corrections for Diagram Color Changes

WoodrowKromar:
#7086 
- Correction to small-screen tabs. 
- Arrows Already functionally in place.
- Allowed to visit the diagram tab when the screen is small, but the diagram is hidden and a warning message is displayed in it's place.

- Corrected the react issues for color changes on diagram nodes.
- Input names are already fixed in size and do not need truncation/cannot replicate.

### DIFF
--- a/src/app/water/water-assessment.service.ts
+++ b/src/app/water/water-assessment.service.ts
@@ -10,7 +10,6 @@ import { DiagramWaterSystemFlows, DischargeOutlet, EdgeFlowData, getComponentNam
   providedIn: 'root'
 })
 export class WaterAssessmentService {
-  isNavigating = false;
   assessmentId: number;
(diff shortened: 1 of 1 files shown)


process-flow-diagram-component/src/components/Drawer/ColorPicker.tsx
```diff
@@ -30,8 +30,10 @@ const ColorPicker = (props: ColorPickerProps) => {
                         name="color"
                         className={'color-input'}
                         value={color}
-                        onChange={(event) => setSelectedColor(event.target.value)}
-                        onBlur={handleSetColor}
+                        onChange={(event) => {
```
nbintertech: Let's remove these changes and skip the requirement for the issue

Issue #7991, opened 2025-11-26
[BUG] Process Heating: Electrotech errors


nbintertech:
**Describe the problem**
Creating new losses in some PHAST electrotech contexts throws errors

**To Reproduce**
Create a new Electrotech PHAST for each of the below

1. Set energy source to electrotech, furnace type to infrared,  add a new Aux Power
2. Set energy source to electrotech, furnace type to EAF, Add a new Exhaust Gas

**Bug 1**
<img width="3793" height="1528" alt="Image" src="https://github.com/user-attachments/assets/44ec058d-220a-4b9c-a80c-c561c45d3162" />

**Bug 2**
<img width="3767" height="1072" alt="Image" src="https://github.com/user-attachments/assets/be363bd3-c179-48d8-b8d5-67961a1339c9" />

nbintertech:
Fixed - Don't calculate Aux Power or EAF on empty user inptu
Fixed - Added validation messaging for AUX Power and EAF
Fixed - Populate default loss name if none is given. Make name required.

Issue #7978, opened 2025-11-21
[BUG] Process Heating - Charge Materials Form Loss Name


nbintertech:
**Describe the problem**
When backspaced or deleted the loss name field seems to reset the name to the default before null is reached

<img width="975" height="448" alt="Image" src="https://github.com/user-attachments/assets/042f2611-4ab4-4b81-a3cb-bee2f6324692" />

**Expected behavior**
1. User should be able to completely delete and rename the field.
2. Verify that other loss (fixture, wall, etc) name fields work correctly. Some appear to function fine, so those can be used as a solution reference. If not, we should fix them.

If the above does not work, or we have a null field value and can't navigate away without error, we can either validate the field or we can populate a default field name at the appropriate time.

**To Reproduce**
Incrementally delete characters from the field

PR #7997, opened 2025-12-01
Issue 7978 7972 - Loss names not saved, various other issues

nbintertech:
#7978 

@rmroot looking for a second set of eyes that I didn't miss losses, other loss views. I will tackle aux power and exhaust gas in #7991


src/app/calculator/furnaces/leakage/leakage-form.service.ts
```diff
@@ -2,6 +2,7 @@ import { Injectable } from '@angular/core';
 import { UntypedFormBuilder, UntypedFormControl, UntypedFormGroup, Validators } from '@angular/forms';
 import { LeakageLoss } from '../../../shared/models/phast/losses/leakageLoss';
 import { GreaterThanValidator } from '../../../shared/validators/greater-than';
+import { V } from '@angular/cdk/scrolling-module.d-ud2XrbF8';
```
rmroot: This doesn't look right..

Issue #7922, opened 2025-10-24
Power Factor Calc Fixes.


WoodrowKromar:
Split from #7907

**Describe the problem**
For months where the power factor exceeds the mandated minimum power factor, the calculator results have negative values for "Demand Penalty" "Penalty Cost" and "Proposed Capacitance" and incorrect values for the "Proposed Reactive Power". Additionally, the proposed fixed capacitance is given as a negative value, where it probably should be zero (with all capacitance being variable capacitance).
{going to ask Chris for help with this one.

<img width="1152" height="457" alt="Image" src="https://github.com/user-attachments/assets/04bc967e-3fbe-4f13-b0ff-d81d14854a5f" />

from Bryan Rasmussen 10/15/25

**Expected behavior**
To be adjusted.

**To Reproduce**
Have the power factor exceed the mandated minimum power factor, which will cause the calculator results to have negative values for "Demand Penalty" "Penalty Cost" and "Proposed Capacitance" and incorrect values for the "Proposed Reactive Power".

WoodrowKromar:
1/14: updates were requested for additional changes.

- Remove the invalid validation when table inputs are at a value of 0.
- The unit on the "Marginal Cost of Demand" should change based on the selection for "How are you billed for demand".
- When screen sizes shrink or the web browser is zoomed in, the table does not flex correctly, becoming cut off until the mobile view takes over.
- Update Proposed Apparent Power units to (kVA) when "How Billed" value is "Apparent Power"
- If "How Billed" value is "Apparent Power", then update table headers "Demand Penalty" and "Penalty Cost" with a prefix of "Excess".

NOTE: Some of the Acceptance criteria has changed in discussions. #8088 tracks the end result and agreed conditions.

PR #8021, opened 2025-12-05
7922 Correction to Calculations

WoodrowKromar:
#7922 
- Adjusted calculations using formula corrections from @pricecr 
- If power factor >= minimum power factor, then the values will be made 0 instead of going negative.
- rest of table results functions corrected to adjust.

### DIFF
--- a/src/app/calculator/utilities/power-factor-correction/power-factor-correction-form/power-factor-correction-form.component.ts
+++ b/src/app/calculator/utilities/power-factor-correction/power-factor-correction-form/power-factor-correction-form.component.ts
@@ -46,9 +46,8 @@ export class PowerFactorCorrectionFormComponent implements OnInit {
     this.form = this.powerFactorCorrectionService.getApparentPowerAndPowerFactor(this.data);
   }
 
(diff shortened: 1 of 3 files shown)


src/app/calculator/utilities/power-factor-correction/power-factor-correction-form/power-factor-correction-form.component.ts
```diff
@@ -38,89 +39,133 @@ export class PowerFactorCorrectionFormComponent implements OnInit {
...
+  constructor(private powerFactorCorrectionService: PowerFactorCorrectionService,
+    private fb: FormBuilder
+  ) {}
 
   ngOnInit() {
+    this.form = this.powerFactorCorrectionService.getApparentPowerAndPowerFactor(this.data);
+  }
+
+  validatePowerFactorInputs() {
+    let monthyInputs = this.form.get('monthyInputs') as FormArray;
```
nbintertech: style: you already have the getter (monthlyInputsFormArray) below to do this

PR #8088, opened 2026-01-16
7922- Additional corrections and adjustments

WoodrowKromar:
#7922 

- Data table validation changed from "value must be greater than 0", to "value must be greater than or equal to 0". This stops the value from erroring at 0, but anything below will trigger the error.
- The "Marginal Cost of Demand" input measurement is updated depending on your selection from "How are you billed for demand".
- Adjusting table input widths to allow table flex, stopping the table from being overlapped while the browser is larger than mobile view, but smaller than full size. (to check, increase zoom on web browser).
- When "Apparent Power (kVA)" is selected for "How are you billed for demand", the unit on the first column of the table is now "kVA".
- When "How Billed" = "Apparent Power (kVA)" update table headers "Demand Penalty" and "Penalty Cost" with a prefix of "Excess". Additionally added notes for results table clarity and check the current math results and calculations.


src/app/calculator/utilities/power-factor-correction/power-factor-correction-form/power-factor-correction-form.component.ts
```diff
@@ -77,6 +77,10 @@ export class PowerFactorCorrectionFormComponent implements OnInit {
   }
 
   setBilledForDemand() {
+console.log('hit billed for demand');
```
nbintertech: Can remove these unless Chris intends to use them for verification, further debugging

src/app/calculator/utilities/power-factor-correction/power-factor-correction-form/power-factor-correction-form.component.html
```diff
@@ -97,77 +99,77 @@
...
+      <ng-container [formGroup]="group">
+        <td>
+          <input type="text" step="any" formControlName="month" id="{{'month_'+i}}" (input)="calculate()" class="table-input-responsive">
+        </td>
+
+        <td
+          *ngIf="(billedForDemand == BilledForDemand.REAL_POWER && adjustedOrActual != AdjustedOrActual.ACTUAL_DEMAND) || billedForDemand == BilledForDemand.APPARENT_POWER">
+          <input type="number" step="any" formControlName="pfAdjustedDemand" id="{{'pfAdjustedDemand'+i}}" (input)="calculate()" (focus)="focusField('pfAdjustedDemand')" class="table-input-responsive">
+            <span *ngIf="group.get('pfAdjustedDemand')?.errors?.required" class="alert-danger pull-right small">Value Required</span>
+            <span *ngIf="group.get('pfAdjustedDemand')?.errors?.greaterThanOrEqualTo == 0" class="alert-danger pull-right small">Value must be ≥ 0</span>
```
nbintertech: Does Chris specifically want the greater than or equal to symbol? Otherwise use the language we use elsewhere for min

src/app/calculator/utilities/power-factor-correction/power-factor-correction-results/power-factor-correction-results.component.html
```diff
@@ -43,14 +48,25 @@
...
+
+      powerFactor - When present is the second input or if 'BOTH' will be the third input
+      names: Power Factor -->
+
       <td *ngIf="inputs.billedForDemand == BilledForDemand.REAL_POWER && inputs.adjustedOrActual == AdjustedOrActual.POWER_FACTOR">
         {{output.realDemand | number:'1.2-2'}}
       </td>
       <td *ngIf="inputs.billedForDemand == BilledForDemand.REAL_POWER && inputs.adjustedOrActual == AdjustedOrActual.ACTUAL_DEMAND">
-       {{output.pfAdjustedDemand | number:'1.2-2'}}
+       wee {{output.pfAdjustedDemand | number:'1.2-2'}}
```
nbintertech: ??

src/app/calculator/utilities/power-factor-correction/power-factor-correction-results/power-factor-correction-results.component.html
```diff
@@ -62,7 +78,8 @@
       </td>
 
       <td>
-        {{output.currentReactivePower | number:'1.2-2'}}
+        <ng-container *ngIf="output.currentReactivePower > 1000000000; else normalReactivePower">&#8734;</ng-container>
```
nbintertech: number is hard to read. Define it as a constant in the component and use "xx_xxx_xxx" syntax

src/app/shared/validators/greater-than-equal-to.ts
```diff
@@ -0,0 +1,26 @@
+import { ValidatorFn, AbstractControl } from '@angular/forms';
+
+
+export class GreaterThanOrEqualToValidator {
```
nbintertech: How is this different than the min validator? "Value can't be less than x" is the same as "Value must be greater than or == x"

nbintertech (changes requested):

src/app/calculator/utilities/power-factor-correction/power-factor-correction-results/power-factor-correction-results.component.html
WoodrowKromar: Actually I need to remove this. Chris and I discussed adding an infinite symbol if the number was too high, but infinite is a real number used by javascript that does appear on our table. if we were to use it if the number became too high, but kept the actual number for calculations, it would be misleading the user.

We ended on not implementing this, so the syntax is not necessary.

nbintertech (approved):

Issue #8060, opened 2026-01-02
Steam Reduction - "Known Variable Field" change not initiating error


koay9f:
From Chris:
Steam reduction, if you change "Known Steam Variable" from "Temperature" to "Quality" the quality field stays at the value for temperature and does not throw the error for "Value can't be greater than 1" red error.
 
~~~~~~~
Was able to mostly reproduce
The field name and unit does change, but the error does not appear.
If you change the value for quality to something else in error, it does appear.

~~~~~~
May also be a good opportunity to transform like the Steam Properties calc (put the "Known Steam Variable" dropdown on the same line as the data entry field)

PR #8125, opened 2026-01-27
8060- Steam Reduction - "Known Variable Field" change

WoodrowKromar:
#8060, #8126 Steam Reduction

- Validation was corrected to the affected inputs based on the selection of Known Variable Inputs. (the input just below, has 4 different labels and uses the same input)
- Corrected logic for Baseline and modification regarding initial inputs. 
- Baseline: removal of base equipment is disabled if base equipment is the only available item. Enables if ther is more than one Baseline.
- Modification: if using "Add Equipment" to add the first Modification, it now defaults to "Add Modified Condition" logic, to stop errors and breaking logic. Functionality is improved.

### DIFF
--- a/src/app/calculator/steam/steam-reduction/steam-reduction-form/steam-reduction-form.component.css
+++ b/src/app/calculator/steam/steam-reduction/steam-reduction-form/steam-reduction-form.component.css
@@ -23,7 +23,7 @@
 }
 
 .disabled {
-    pointer-events: none;
-    opacity: 0.5;
-    cursor: not-allowed;
+  pointer-events: none;
+  opacity: 0.5;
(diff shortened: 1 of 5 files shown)


nbintertech (changes requested):
The 2 comments "Good.." are *spoiler* good, and should not be changed. 

I can't see a need for the rest of the changes made. We should be able to simply disallow deleting the baseline by conditionally rendering the delete button. EX.. display delete button "if not in baseline OR we are in baseline and is NOT  first item". See the compressed air reduction calculator. This should match with what I mentioned in slack the other day.

nbintertech (changes requested):

nbintertech (approved):

Issue #8111, opened 2026-01-22
Hide non-TMY3 data in Weather Select Location


nbintertech:
**Describe the problem**
Locations with non TMY3 data are showing up as available, being selected, leading to errors

**Expected behavior**
Users should only see station options with TMY3 data

**To Reproduce**
Visit System Information --> Weather Tab and enter Oakridge zipcode.

Issue #8093, opened 2026-01-20
[BUG] Process Cooling: Weather Data bugs


nbintertech:
**Describe the problem**
Weather data seems to be working only intermittently, QA users experience a handful of different sideffects. Will add more when I investigate

**Expected behavior**

**To Reproduce**

**Additional context**
Observed during Process Cooling QA meeting

nbintertech:
Note from @pricecr re: world map on weather search -> if we stick with our US only TMY data we don't really need the world map. Because this uses a borrowed maps API and set of coordinates, may not be simple to change. Investigate.

nbintertech:
Quick change from Riyaz - use line instead of line_marker on weather chart

PR #8127, opened 2026-01-27
Issue 8093 8011 - Update Process Cooling Weather workflow

nbintertech:
#8093 #8111 

A few things that are not part of the issue reqs:
- Updates a shared 'confirm actions' modal
- Use better NG mechanics for routerLink li/a tags, "@let" etc
- Add documentation for 'weather-data' module pulled from VERIFI


src/app/shared/modules/weather-data/weather-data.component.html
WoodrowKromar: https://weather.ornl.gov/api/  webpage does not properly work when searching for a location using Explore data:

image

Error: ERROR TypeError: Cannot read properties of undefined (reading 'lat')


Code functionality works well, so not necessarily something to fix in this PR.

WoodrowKromar (approved):

Issue #8075, opened 2026-01-08
Email MEASUR Assessment: Modify filename, multiple receivers


nbintertech:
- We want to add functionality to name the files we send.
- Allow multiple receivers

nbintertech:
Note for us: nodemailer (weather api smtp lib) only supports passing in comma-separated addresses. We can do some cleaning on the backend if team/users want to support semicolons as well.

PR #8138, opened 2026-01-30
8075 - Allows Multiple Email Recipients

WoodrowKromar:
#8075 

- adding validation for multiple email receivers. 
- nodemailer handles multiple emails in strings with comma separations, or an array. either works.


nbintertech (approved):
Looks good. Please hold on merging for now, I need to look into weather api changes

Please add placeholder text below the field "Separate multiple email addresses with commas (ex. user1@example.com, admin@example.com)"

Issue #7869, opened 2025-10-06
Custom Materials: Lighting Fixtures


rmroot:
Lighting fixtures have been added to the suite: https://github.com/ORNL-AMO/MEASUR-Tools-Suite/issues/15

@koay9f can elaborate but the desire is for lighting fixtures to be treated like other custom materials, where users can create there own. This issue has two parts, possibly three, that can be broken up into different issues.

1. Migrate the fixtures db to use the suite and remove the current .csv
2. Add custom fixtures to the custom materials section (probably re-name)
3. It may be worth giving that whole section a facelift

* flue gas materials leverage behavior subjects. It may be worth the effort to implement behavior subjects for the other materials as well. Update this issue and create issues as needed.

PR #8141, opened 2026-01-31
7869 - Lighting Fixtures Step 1

WoodrowKromar:
#7869 Step 1
- Not actually about custom materials, about Migrate the fixtures db to use the suite. 
- No .csv files, setting in lighting-data as an object.
- New service connecting to DB for default data.
- still parsing data into an object as no LightingFixtureCategories data or interface present.
- integrated into lighting-replacement.


nbintertech (changes requested):
We don't need to call getLightingSystems in multiple components:
When the top-level LR component initializes, call a LR service method to set a service state for lightingFixtureCategories (retrieved from the api service). The form and help components can access lightingFixtureCategories from that service state.

nbintertech (approved):

Issue #7883, opened 2025-10-10
Treasure Hunt Sort & Filter Enhancement


koay9f:
**Describe the feature**
1.	Filter by system in the “Find Treasure” area of the treasure hunt module
2.	Sort “Treasure Chest” by difficulty to implement
From Chris Price 4/17/25

**Describe MEASUR location**
MEASUR Treasure Hunt

**Describe MEASUR UI**
1. see other issue
2. We have one for utility type, but "Electricity" covers almost all calcs, so if you just want "compressed air" calcs, this is not very useful.

<img width="497" height="216" alt="Image" src="https://github.com/user-attachments/assets/67669c2c-650d-461e-a1f7-298daeb790d6" />

3. Add "Effort" (as in "Effort to Implement") to the sort list here:

<img width="711" height="289" alt="Image" src="https://github.com/user-attachments/assets/ea63579d-b397-4d3c-9b64-2465eef8c4d3" />
 

4. 2b. Originally I thought 2 was for Treasure Chest, not "Find Treasure". However, if we are looking at the sort/filter list of the chest anyway... Given that there are only 2 other things in "Add'l Filters" - maybe we should either have those two things be in the regular dropdown OR move all filters to the modal. 

<img width="680" height="157" alt="Image" src="https://github.com/user-attachments/assets/f7683662-a7a2-4b92-bd78-824143fe8974" />

<img width="772" height="518" alt="Image" src="https://github.com/user-attachments/assets/333db6e5-d9dc-41ec-a4e5-088cacd936f4" />

 
**Who should verify**Kristina
---
### Optional
**Feature requested by**
Chris Price

nbintertech:
Edited to remove feedwater economizer requirement. @koay9f what is the other issue - "see other issue"?

koay9f:
Point 3 - Filter by "Effort" still needs to be included
Point 2 - do not need the "Other/none" entry in dropdown, there are not calcs in it at this time
I like all the fixes, just a few little things to tidy up.  These should not delay the patch release, what is in is okay

PR #8159, opened 2026-02-06
7883 - Sort By Implementation Effort

WoodrowKromar:
#7883 

- On Treasure Chest filter and sort, you can now sort by Implementation Effort.


nbintertech (changes requested):
To my knowledge the the TH opportunities don't already sort by Implementation Effort. I would expect we're going to need more implementation than just adding a sortBy string here. We can also see this isn't working in it's current state by looking at the first three cards sorted in the TH example.

nbintertech (approved):

Issue #7406, opened 2025-03-20
Diagram: indicate states on the component


nbintertech:
**Describe the feature**
From the global diagram view, the user should be able to see that *something* is wrong with the component (invalid field, "Estimated Unknown Loss" notification) by looking at it. 

Visibility of this state should have the option of being turned off globally.

**Describe MEASUR UI**
1. Implement an icon in the style of existing validation (circle/exclamation, triangel/exclamation) over components that have invalid field, or estimated unknown loss states. 
2. The user should be able to turn off the visibility of these icons in the Alerts tab

<img width="57" height="52" alt="Image" src="https://github.com/user-attachments/assets/a0549db5-0ab5-47cc-ac68-16a1f5f54615" />

<img width="58" height="51" alt="Image" src="https://github.com/user-attachments/assets/c7bc2c11-170c-46a3-bd3a-e6e3126d7d8e" />

---
### Optional
**Feature requested by**
Tom W, Kristina, Kiran

PR #8215, opened 2026-02-24
7406 - Icons added to Water Diagram

WoodrowKromar:
#7406 
- Icons present on Node Edit forms are no present directly on the Diagram nodes themselves.


process-flow-diagram-component/src/App.css
```diff
@@ -378,6 +379,19 @@ a {
 .react-flow__node.selectable.selected .custom-node-toolbar {
   visibility: visible;
   pointer-events: all;
+} */
+ 
+/* Show the Toolbar edit button only on hover */
```
nbintertech: The edit button was intentionally shown constant if selected so there was better indication about which component was being edited in the data drawer. If Kiran/Kristina don't care about keeping that, all good

Issue #8210, opened 2026-02-23
Notes


nbintertech:
**Describe the feature**
Users would like to be able to add notes to the diagram. Some MEASUR assessments already let users add notes. Familiarize yourself with these implementations. See equipment notes and loss notes in compressed air and PHAST.

Ensure notes export with a diagram or given assessment

**Describe MEASUR location**
Assumption is that there is a single note set scoped to the diagram and connected assessment if there is one

**Describe MEASUR UI**
If it's scoped as assumed above, implement in the left diagram sidebar. Create a new tab for notes. 
Stylize like other MEASUR note implementations

**Unknowns**
Scope of notes (diagram as a whole? per component?)
Should they also show in the assessment? only in the assessment? only in the diagram?

**Who should verify**
Kiran/Kirstina

---
### Optional
**Feature requested by**
Users, Kiran through post-training feedback

PR #8262, opened 2026-03-11
8210 - Water Diagram Notes

WoodrowKromar:
#8210

- Notes added to Water diagram
- Chevrons added to cycle left and right through left drawer tabs
- Water Diagram left drawer tabs reordered so help is the last tab, and Notes and Alerts are visible initially
- Added to the already present state. Was initially its own state but I corrected to be coupled with the rest of water diagram


nbintertech (changes requested):
Can we display: none the chevron to the left of "Build". It's already being hidden. This would nicely free up tab space
Image

nbintertech (changes requested):
We still have the hidden chevron taking up useful space on the left side of the tabs. Would be great to have this space back.

Image

process-flow-diagram-component/src/components/Drawer/MenuSidebar.tsx
```diff
@@ -336,6 +342,35 @@ const MenuSidebar = memo((props: MenuSidebarProps) => {
...
         <TabPanel value={selectedTab} index={3}>
+          <Box sx={{ flexGrow: 1, paddingY: '1rem', paddingX: '.5rem' }}>
+            <div className="form-group pt-4">
+              <TextField
+                id="diagramNotes"
+                name="diagramNotes"
+                label="Diagram Notes"
+                multiline
+                minRows={8}
+                value={diagramNotes}
```
nbintertech, line 353: empty string should be passed if null/undefined
Image

nbintertech (changes requested):

nbintertech (approved):

Issue #8146, opened 2026-02-03
Custom Materials: Lighting Fixtures Component


WoodrowKromar:
**Describe the feature**

@koay9f can elaborate but the desire is for lighting fixtures to be treated like other custom materials, where users can create there own. This issue has two parts, possibly three, that can be broken up into different issues.

* flue gas materials leverage behavior subjects. It may be worth the effort to implement behavior subjects for the other materials as well. Update this issue and create issues as needed.

Custom materials lacks lighting fixtures as an option. Add custom fixtures to the custom materials section (probably re-name).
It may be worth giving that whole section a facelift. 

**Describe MEASUR location**
custom-materials

**Describe MEASUR UI**
Add new section and lighting-fixtures component to handle data type. Cannot be generic.

**Who should verify**
Nick

PR #8309, opened 2026-03-24
8146 - Custom Materials: Lighting Fixtures

WoodrowKromar:
#8146 
Adds a new section of Custom Materials: Lighting Fixtures.
- Functions the same as other materials, allowing CRUD operations and saved in the Indexdb service.
- Allows user to select the Material, then the type of their fixture.
- Help Text added for individual field entries.

COMPLETE


nbintertech (changes requested):
The branch doesn't build for me. I tried to run a development builld from an existing install, and also tried to reset, install, run development build.

image

src/app/suiteDb/lighting-fixtures/lighting-fixtures-material.component.css
```diff
@@ -0,0 +1,8 @@
+.difference-error{
```
nbintertech: Definition doesn't appear to be used

src/app/suiteDb/lighting-fixtures/lighting-fixtures-material.component.html
```diff
@@ -0,0 +1,108 @@
...
+                        <div class="col-12">
+                            <div class="header card-header">
+                                <h3>Confirm Delete</h3>
+                            </div>
+                        </div>
+                    </div>
+                    <div class="delete-warning-container">
+                        <div class="row">
+                            <div class="col-12">
+                                Fixture Type: {{newMaterial.type}}
```
nbintertech: Why use type here in contrast to the other components which display the name?

src/app/suiteDb/custom-materials/custom-materials.service.ts
```diff
@@ -78,9 +83,17 @@ export class CustomMaterialsService {
...
       this.importWallLossSurfaces(data.wallLossesSurface);
     }
+    if (data.lightingFixtureMaterial.length !== 0) {
+      this.importLightingFixtures(data.lightingFixtureMaterial);
+    }
   }
 
 
+  async importLightingFixtures(data: Array<LightingFixtureMaterial>) {
+    //flesh this out. for now just set selectedLightingFixtures to data
```
nbintertech: We should be able to import and export the lighting fixtures

src/app/suiteDb/custom-materials/custom-materials.component.html
```diff
@@ -133,6 +133,22 @@
...
 
+<div class="col-11 mx-auto mb-5">
+  <div class="tab-content content-container">
+    <div class="card-form">
+      <label class="group-label general" (click)="toggleMaterial('lighting-fixtures')">Lighting Fixtures
+        <span class="fa fa-chevron-up pull-right m-1" *ngIf="!showLightingFixturesMaterials"></span>
+        <span class="fa fa-chevron-down pull-right m-1" *ngIf="showLightingFixturesMaterials && hasLightingFixtureMaterials"></span>
+      </label>
+      <a class="click-link pull-right small" *ngIf="showLightingFixturesMaterials" (click)="toggleModal('lighting-fixtures')">+Add New Lighting Fixture</a>
+      <!-- <p class="help-text">View, create, edit your custom lighting fixtures.</p> -->
```
nbintertech: Remove comment

src/app/calculator/lighting/lighting-replacement/lighting-replacement.component.ts
```diff
@@ -54,7 +54,7 @@ export class LightingReplacementComponent implements OnInit {
     private analyticsService: AnalyticsService) { }
 
   ngOnInit() {
-    this.lightingReplacementService.setLightingFixtureCategories();
+    // this.lightingReplacementService.setLightingFixtureCategories();
```
nbintertech: Remove comment

src/app/calculator/lighting/lighting-replacement/lighting-replacement.service.ts
```diff
@@ -95,9 +90,11 @@ export class LightingReplacementService {
   }
 
   generateExample(isBaseline: boolean): LightingReplacementData {
+
+    const lightingFixtureCategories = [];
```
nbintertech: This array will never have items

src/app/calculator/lighting/lighting-replacement/lighting-replacement.service.ts
```diff
@@ -17,17 +17,12 @@ export class LightingReplacementService {
   operatingHours: OperatingHours;
   lightingFixtureCategories: Array<LightingFixtureCategory>;
```
nbintertech: This is never set and is still being accessed

src/app/tools-suite-api/lighting-suite-api.service.ts
```diff
@@ -1,19 +1,22 @@
 import { Injectable } from '@angular/core';
 import { ToolsSuiteApiService } from './tools-suite-api.service';
-
 @Injectable({
     providedIn: 'root'
 })
 export class LightingSuiteApiService {
+
```
nbintertech, line 8: I don't understand why the changes to this service were made. We're still calling this method in both the replacement module and the custom materials module. This now also mixes getter/setter patterns. If the objective is load all materials a single time on app load so that they can be used in custom materials and replacement, we should do that in core.component/service. I think the simplest solution is to discard changes to this service and stick with what we had

nbintertech (changes requested):
Implementation is largely in line with the other material module components

We should just hide fixture type if it's unused in the 'Custom' context.
Image

src/app/suiteDb/lighting-fixtures/lighting-fixtures-material.component.html
WoodrowKromar: Because the type and name are not the same. The Type field has static selectable values that determine the lighting fixture's data. I added a name field below and added name to the interface so users can enter names. The name is the field at the top of the custom material when viewing all materials for that section.

src/app/tools-suite-api/lighting-suite-api.service.ts
```diff
@@ -120,8 +120,4 @@ export interface LightingFixtureData {
     lumenDegradationFactor: number
 }
 
-export interface LightingFixtureCategory {
```
nbintertech, line 123: This was good - please restore. Having the defined interface is preferred to an object type that must be copied/pasted around the file/app. There is a blip in the style guide

src/app/tools-suite-api/default-compressor-api.service.ts
```diff
@@ -11,6 +11,7 @@ export class DefaultCompressorApiService {
     ) { }
 
     getGenericCompressors() {
+                console.log('comp', this.toolsSuiteApiService.ToolsSuiteModule);
```
nbintertech: Remove if not for extended debugging

src/app/suiteDb/lighting-fixtures/lighting-fixtures-material.component.html
nbintertech: Rephrasing comment to be more clear: In the name of good UX, when we are rendering a delete confirmation for the user's material, the name should be displayed instead of the type.

src/app/suiteDb/custom-materials/custom-materials.service.ts
```diff
@@ -90,8 +90,14 @@ export class CustomMaterialsService {
 
 
   async importLightingFixtures(data: Array<LightingFixtureMaterial>) {
-    //flesh this out. for now just set selectedLightingFixtures to data
-    this.selectedLightingFixtures = data;
+    for (let i = 0; i < data.length; i++) {
```
nbintertech, line 93: Good - it looks like we also need to update the material listing in the export window:

Image

nbintertech (changes requested):

nbintertech (approved):

Issue #8314, opened 2026-03-24
Water Diagram: Remove Total Discharge validation on Water Using Systems


nbintertech:
**Describe the feature**
The validation for value > 0 must be removed for Total Discharge Flow under Water Using System components only. This is to allow for typical situations where all water may be used in product. We should no longer see the form error, nor indication on the diagram component

Validation should still ensure a value is not less than 0 if entered
Individual flow validation should also match the above

**Describe MEASUR location**
Data Drawer with water-using-system type as active component/node

---
### Optional
**Feature requested by**
Kiran

nbintertech:
Enhanced: Water using systems may decide to use all water in product and/or known losses without receiving validation or warning messages.

<img width="1686" height="1682" alt="Image" src="https://github.com/user-attachments/assets/ae848403-6e6c-4d3e-98e7-dbe30266e164" />

PR #8318, opened 2026-03-25
8314 - Validation Adjustment

WoodrowKromar:
#8314 Validation adjusted 

- Only on Outflow for water using systems.
- Outflow only errors "Total Discharge Flow must be greater than or equal to 0 for water using systems" if selected node is a water using system and the total discharge is less than 0.
- Other error functionality is unchanged.

### DIFF
--- a/process-flow-diagram-component/src/validation/Validation.ts
+++ b/process-flow-diagram-component/src/validation/Validation.ts
@@ -3,7 +3,7 @@ import { DiagramSettings, validateKnownLosses, validateTotalFlowValue } from "pr
 import * as Yup from 'yup';
 export const TOTAL_SOURCE_FLOW_GREATER_THAN_ERROR = `Total Source Flow must be greater than 0`;
 export const TOTAL_DISCHARGE_FLOW_GREATER_THAN_ERROR = `Total Discharge Flow must be greater than 0`;
-export const TOTAL_DISCHARGE_FLOW_GREATER_THAN_OR_EQUAL_TO_ZERO_ERROR = `Total Discharge Flow must be greater than or equal to 0 for water using systems`;
+export const TOTAL_DISCHARGE_FLOW_GREATER_THAN_OR_EQUAL_TO_ZERO_ERROR = `Total Discharge Flow must be greater than or equal to 0`;
 // todo move to forms
 const getSystemNumberFieldValidation = (fieldLabel: string) => Yup.number()
     .nullable()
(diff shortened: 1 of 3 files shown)


process-flow-diagram-component/src/validation/Validation.ts
```diff
@@ -3,7 +3,7 @@ import { DiagramSettings, validateKnownLosses, validateTotalFlowValue } from "pr
 import * as Yup from 'yup';
 export const TOTAL_SOURCE_FLOW_GREATER_THAN_ERROR = `Total Source Flow must be greater than 0`;
 export const TOTAL_DISCHARGE_FLOW_GREATER_THAN_ERROR = `Total Discharge Flow must be greater than 0`;
-
+export const TOTAL_DISCHARGE_FLOW_GREATER_THAN_OR_EQUAL_TO_ZERO_ERROR = `Total Discharge Flow must be greater than or equal to 0 for water using systems`;
```
nbintertech: Can omit 'for water using systems'

process-flow-diagram-component/src/validation/Validation.ts
```diff
@@ -35,36 +35,48 @@ export const getEstimateSystemValidationSchema = (
...
+    sumUserKnownLosses?: number,
+    isWaterUsingSystem?: boolean
+): Yup.ObjectSchema<FlowForm> => {
+    const totalFlowError = flowLabel === 'Source' ? TOTAL_SOURCE_FLOW_GREATER_THAN_ERROR : (isWaterUsingSystem ? TOTAL_DISCHARGE_FLOW_GREATER_THAN_OR_EQUAL_TO_ZERO_ERROR : TOTAL_DISCHARGE_FLOW_GREATER_THAN_ERROR);
     const unit = settings.unitsOfMeasure === 'Imperial'? 'Mgal' : 'm<sup>3</sup>';
+    let totalFlowSchema = Yup.number().nullable().test(
+        'total-flow-min',
+        totalFlowError,
+        function (value) {
+            if (value === null || value === undefined) return true;
```
nbintertech: I believe nullable already covers this conditional but you should check documentation

process-flow-diagram-component/src/validation/Validation.ts
```diff
@@ -35,36 +35,48 @@ export const getEstimateSystemValidationSchema = (
...
+    );
+
+    totalFlowSchema = totalFlowSchema.test(
+        'sum-differs',
+        totalFlowError,
+        function (value) {
+            const { path, createError } = this;
+            const flowDifference = Math.abs(totalCalculatedFlow - value);
+            const unallocated = Number((flowDifference).toFixed(settings.flowDecimalPrecision));
+            const totalCalculatedFlowFixed = Number(totalCalculatedFlow.toFixed(settings.flowDecimalPrecision));
```
nbintertech: If no longer used, remove please

process-flow-diagram-component/src/validation/Validation.ts
```diff
@@ -35,36 +35,48 @@ export const getEstimateSystemValidationSchema = (
...
+            if (flowLabel === 'Discharge' && isWaterUsingSystem) {
+                return value >= 0;
+            }
+            return value > 0;
+        }
+    );
+
+    totalFlowSchema = totalFlowSchema.test(
+        'sum-differs',
+        totalFlowError,
```
nbintertech: Replace totalFlowError with `Total Flow must be equal to the sum of individual flows` as placeholder.

totalFlowError is a placeholder that's overridden by the message returned from createError. It will also reflect the value assignment above which does not have to do with sum-differs validation.

nbintertech (changes requested):
We missed this case in requirements. The individual flow should also allow nullable/0 value to align with the total flow field, otherwise we'll see the below error.

Image

process-flow-diagram-component/src/validation/Validation.ts
WoodrowKromar: Nullable just accepts the null value and stops typeOf errors that can be caused if a null is entered. The if statement stops the test from running, preventing rest of functionality from being run. Nullable does not stop this.

nbintertech:
Adapted and merged in a different PR

Issue #7560, opened 2025-05-28
All Plotly Graphs - modebar


koay9f:
Be sure plotly interactive controls / modebar are on

Remove Box Select and Lasso Select EDIT: Alex is going to explore this the next INPLT so can pause for now
Remove plotly logo 

pies and bars do not need the zoom features (not sure if they even apply...)

pay special attention to graphs in Data Explorer

See if `drawline`, `drawopenpath`, `drawcircle`, `drawrect`, `draseshape` work in javascript plotly
I found them here: https://plotly.com/python/configuration-options/
that is for python. the javascript version doesn't have it explicitly called out  but they are in the "2D shape drawing" `Remove Modebar Buttons` section

This could be really helpful in Data Explorer HOWEVER, it is possible it would be too much for it to handle so let's experiment.  
I'm not sure where else it would be useful and if this would be better moved to another issue, make a note and do so please. 

INFORMATION UPDATE: Acceptance Criteria updated
- Remove All Zoom Features from Pie and Chart graphs
- Remove Box and Lasso Select from every plotly graph instance as well as the logo.
UPDATED CRITERIA 4/13:
- Dot and Line graphs must have zoom functionality

koay9f:
Remove Box Select and Lasso Select - right now these are only in Data Explorer>Day Types I think and that is not actually a good use case for them. Let's remove them from here and let me know any other place they are around.
I think Box select is in "Data Visualization", let's leave that one for now and let Alex explore this the next INPLT

nbintertech:
We'll break this down into smaller pieces and target high use areas. @WoodrowKromar  let's chat on this before you begin.

WoodrowKromar:
UPDATED CRITERIA:
- Remove zoom from ALL Pies and Bar charts. 
- Remove all logos from every graph if possible.
- Drawling and other draw functionalities DO work. Tested in VERIFI.
- Remove box and lasso select from all plotly graphs.
- there are no exceptions. It should removed from all of them.

WoodrowKromar:
Updated Criteria: Dot and Line graphs (scatter) must have zoom functionality

PR #8346, opened 2026-04-07
Issue 7560

WoodrowKromar:
#7560
- Zoom Functionality removed from all 'Pie' and 'Bar' graphs.
- Some Plotly graphs were missing default config. It was corrected.
- Default Config helper function adjusted to handle new functionality.

# Pull Request Template

## Checklist
- Variable, function, and class names are descriptive
- Code leverages existing architecture, helper methods, and shared modules when applicable
- Development artifacts such as console logs, test values, and comments have been removed
- All Copilot, Claude Code, or other LLM implementations have been reviewed as if contributed by another developer
- UI changes have been manually tested for usability and appearance
- This PR references the related issue number (if applicable), ex. issue -> `#0000`

## Description
- **For development team**: Optional. Update the issue with a comment, if necessary.
- **For open source contributors**: The PR description clearly explains the purpose and scope of the changes

---

### DIFF
--- a/src/app/shared/helperFunctions.ts
+++ b/src/app/shared/helperFunctions.ts
@@ -48,7 +48,9 @@ export function getRandomFlatColor(): string {
 
 
 // Returns a Plotly config object with base options, merging any provided config
-export function defaultPlotlyConfig(config?: Partial<SimpleChart['config']>): Partial<SimpleChart['config']> {
+export function defaultPlotlyConfig(config?: Partial<SimpleChart['config']>, chartType?: string): Partial<SimpleChart['config']> {
+
+    console.log(config);
     const mergedConfig = {
         modeBarButtonsToRemove: ['select2d', 'lasso2d'],
         displaylogo: false,
@@ -58,3 +60,20 @@ export function defaultPlotlyConfig(config?: Partial<SimpleChart['config']>): Pa
     };
     return mergedConfig;
 }
+
+/**
+ * Returns the first matching Plotly trace type from the provided array, or 'scatter' if none found.
+ * @param traces Array of trace objects (must have a 'type' property)
+ * @param typesToFind Array of type strings to look for (e.g., ['bar','scatter'])
+ */
+// export function getFirstTraceTypeOrDefault(traces: Array<{ type?: string }>, typesToFind: string[] = ['scatter','bar','line','pie','sankey','scattergl']): string {
+//     if (!Array.isArray(traces)) return 'scatter';
+//     for (const trace of traces) {
+//         if (trace && typeof trace.type === 'string' && typesToFind.includes(trace.type)) {
+//             return trace.type;
+//         }
+//     }
+//     return 'scatter';
+// }
+
+export function ()
--- a/src/app/shared/helperFunctions.ts
+++ b/src/app/shared/helperFunctions.ts
@@ -56,29 +56,19 @@ export function getRandomFlatColor(): string {
  */
 export function defaultPlotlyConfig(config?: object, chartType?: string | unknown): object {
     let modeBarButtonsToRemove = ['select2d', 'lasso2d'];
-    let typeToCheck: any = chartType;
-
-    if (!typeToCheck && config) {
-        if ((config as any).type !== undefined) {
-            typeToCheck = (config as any).type;
-        } else if (Array.isArray((config as any).data)) {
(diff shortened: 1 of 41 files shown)


src/app/shared/helperFunctions.ts
```diff
@@ -48,9 +48,16 @@ export function getRandomFlatColor(): string {
 
 
 // Returns a Plotly config object with base options, merging any provided config
-export function defaultPlotlyConfig(config?: Partial<SimpleChart['config']>): Partial<SimpleChart['config']> {
+export function defaultPlotlyConfig(config?: Partial<SimpleChart['config']>, chartType?: string | any): Partial<SimpleChart['config']> {
```
nbintertech: Good solution. 

1 Issue - Note that if chartType is passed as string, the modeBarButtonsToRemove is not affected. I'd prefer to keep string in the optional type.

Suggestions:
- Replace `any` with `unknown` <- better TS convention for this case 
- Replace the typing of `Partial<SimpleChart['config']>`  to simply `object`. I'm doubting how consistently SimpleChart is used throughout the app, among other issues.
- In the future consider jsdoc instead of the comment above the method sig, to better explain what expected shapes chartType could look like. Search app for `@param` for examples.

nbintertech (changes requested):

nbintertech (approved):

src/app/shared/helperFunctions.ts
```diff
@@ -57,8 +57,12 @@ export function getRandomFlatColor(): string {
 export function defaultPlotlyConfig(config?: object, chartType?: string | unknown): object {
```
nbintertech, line 57: See original comment - nothing happens if chartType is passed as a string

nbintertech (changes requested):

src/app/shared/helperFunctions.ts
```diff
@@ -56,13 +56,31 @@ export function getRandomFlatColor(): string {
  */
 export function defaultPlotlyConfig(config?: object, chartType?: string | unknown): object {
     let modeBarButtonsToRemove = ['select2d', 'lasso2d'];
-    if (Array.isArray(chartType)) {
-        // Remove zoom for pie and bar
-        if (chartType.some(trace => ['pie', 'bar'].includes(trace.type))) {
-            modeBarButtonsToRemove.push('zoom2d', 'zoomin2d', 'zoomout2d');
+    let typeToCheck: any = chartType;
+
+    if (!typeToCheck && config) {
```
nbintertech: Why is this block needed and why are we now treating the config as a potential chartType (i.e. traces)? This is confusing and we shouldn't do it.

From the commit before this, we just needed an else to check typof === string and run associated logic to satisfy the comment left.

nbintertech (changes requested):

src/app/shared/helperFunctions.ts
WoodrowKromar: The reason for using the config now is because some calls don't send a chart type separately, the type is defined in the config as a whole. These are getting bypassed by the function because it is only checking the second parameter: chart type.

export function defaultPlotlyConfig(config?: object, chartType?: string | unknown): object {


1. this.plotlyService.newPlot(this.o2UtilizationChart.nativeElement, [trace], layout, defaultPlotlyConfig(config));
2. this.plotlyService.newPlot(this.powerFactorTiangle.nativeElement, data, layout, defaultPlotlyConfig(modebarBtns));
3. this.plotlyService.newPlot(this.compressedAirInventoryBarChart.nativeElement, dataArray, layout, defaultPlotlyConfig(configOptions));

nbintertech: I can't find any examples of a caller that is passing `config` that has a type i.e. `bar`, `scatter`. Even so, the method sets the rules for passed args and callers should cater to that instead of vice versa. Isn't this the exact situation we introduced a chartType optional type?

In the interest of time I'd be inclined to approve despite the above, but there is another issue with the changes made to `config`. `config` shouldn't have a `.data` array and so is error prone. If it does, it is not `config` and the wrong argument is being passed. Either greater clarity needs to be made about what `object` should be or we should go back to an explicit type (the Partial) of SimpleChart. 

Lets cut losses here. Implement the minimum possible fix to avoid errors and we can move on.

nbintertech (approved):

Issue #8394, opened 2026-04-28
Finish Process Cooling banner links: Share data, Privacy disclosure


nbintertech:
This should be done after #8359 is merged.

**Describe the feature**
Implement the share data and privacy/security disclosure buttons. These should be done separately -> please create a new issue for one of them 

- Use the angular dialog cdk workflow that's part of modal-dialog service, instead of the legacy MEASUR pattern which uses the boostrap modal, modalOpen subscriptions, and including the modal component as a child in the html.
- If the markup for the existing modal isn't decoupled from the bootstrap modal, duplicate the markup to a new component like what was done for commit 5a08955d43c2334653ee6370f8d571e11703f4f7 of PR
https://github.com/ORNL-AMO/AMO-Tools-Desktop/pull/8393
- Make frequent pulls from `process-cooling` and integrate them into your branch by first stashing changes or rebasing your branch commits. This will limit conflicts on PR merge.

Rely on above PR as a conceptual guide for what we need to do.

**Describe MEASUR location**

<img width="203" height="51" alt="Image" src="https://github.com/user-attachments/assets/827ff3ce-91ac-4747-9866-fb0d2189a42b" />

PR #8432, opened 2026-05-11
Process cooling issue 8394 w

WoodrowKromar:
https://github.com/ORNL-AMO/AMO-Tools-Desktop/issues/8394 - for only Process Cooling

Create a similar Privacy component with decoupled styling from all Bootstrap/modal classes.
Uses the CDK Routing as a new standard (currently only present in Process Cooling).
https://github.com/ORNL-AMO/AMO-Tools-Desktop/pull/8420 - for only Process Cooling

Create a similar Email component with decoupled styling from all Bootstrap/modal classes.
Uses the CDK Routing as a new standard (currently only present in Process Cooling).

### DIFF
--- a/src/app/shared/import-export/export-item/export-item.component.ts
+++ b/src/app/shared/import-export/export-item/export-item.component.ts
@@ -1,4 +1,4 @@
-import { Component, EventEmitter, inject, Inject, Input, Output, ViewChild } from '@angular/core';
+import { Component, inject, Inject, Input, ViewChild } from '@angular/core';
 import { DirectoryDashboardService } from '../../../dashboard/directory-dashboard/directory-dashboard.service';
 import { DirectoryDbService } from '../../../indexedDb/directory-db.service';
 import { Assessment } from '../../models/assessment';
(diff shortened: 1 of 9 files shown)


src/app/shared/import-export/export-item/export-item.component.ts
```diff
@@ -58,9 +56,10 @@ export class ExportItemComponent {
   hideExportModal() {
     this.exportService.exportAll = false;
     if (this.inAssessment) {
-      this.close.emit(false);
+      this.dialogRef.close('false');
```
nbintertech: We don't need to pass a close result value for either of these calls

src/app/shared/security-and-privacy/security-and-privacy-item/security-and-privacy-item.component.ts
```diff
@@ -0,0 +1,17 @@
...
+    templateUrl: './security-and-privacy-item.component.html',
+    styleUrls: ['./security-and-privacy-item.component.css'],
+    standalone: false
+})
+export class SecurityAndPrivacyItemComponent {
+    dialogRef = inject<DialogRef<string>>(DialogRef);
+
+
+    hidePrivacyModal() {
+        this.dialogRef.close('false');
```
nbintertech: We don't need to pass a value. Just close()

nbintertech (changes requested):
Please add a note to the top of each of those new modal components so we know they're meant to  work with the global modal and replace the existing Bootstrap implementation

nbintertech (approved):

Issue #8440, opened 2026-05-13
Chiller Y axis max must be calculated instead of static


WoodrowKromar:
For the Chiller performance chart y axis - it is statically set to 1.2.. We should be able to just ask copilot to derive the axis max from max of the efficiency value. If things go haywire just manually calculate it with some buffer percentage increase from the max.

**Describe the feature**
Describe new functionality from the user perspective if possible

**Describe MEASUR location**
Include assessment or calculator location, field names, context details to reach the relevant state

**Describe MEASUR UI**
Outline UI ideas, refer to an existing MEASUR pattern, or Dev's choice

**Who should verify**

---
### Optional
**Feature requested by**

**Sample Inputs and Results**
Include input testing values and expected results

**Future Integrations**
Future connections to existing MEASUR features, VERIFI, JUSTIFI, etc

Issue #8439, opened 2026-05-13
Add Design to PC Outdoor Air Temperature


WoodrowKromar:
Change "Outdoor Air Temperature" to Design Outdoor Air Temperature" in System Information - > Condenser Cooling if the cooling method is "Air"

<img width="1911" height="219" alt="Image" src="https://github.com/user-attachments/assets/7ab66b9b-969e-4f91-ad08-6de05ac74e25" />

**Describe the feature**
Describe new functionality from the user perspective if possible

**Describe MEASUR location**
Include assessment or calculator location, field names, context details to reach the relevant state

**Describe MEASUR UI**
Outline UI ideas, refer to an existing MEASUR pattern, or Dev's choice

**Who should verify**

---
### Optional
**Feature requested by**

**Sample Inputs and Results**
Include input testing values and expected results

**Future Integrations**
Future connections to existing MEASUR features, VERIFI, JUSTIFI, etc

PR #8442, opened 2026-05-13
Issue 8439, 8440 -  Process Cooling: Update Air Design Temp label, Perf chart Chart styling

WoodrowKromar:
#8439
- Changed "Outdoor Air Temperature" to Design Outdoor Air Temperature" in System Information - > Condenser Cooling if the cooling method is "Air"

#8440 
- Added Calculated Max values for y axis.
- Maintaining a maximum of 7 value measurement ticks that also adjust depending on the values chosen.

### DIFF
--- a/src/app/process-cooling-assessment/services/process-cooling-charts.service.ts
+++ b/src/app/process-cooling-assessment/services/process-cooling-charts.service.ts
@@ -28,6 +28,11 @@ export class ProcessCoolingChartsService {
         hovertemplate: `${chiller.name}<br>Load: %{x}<br>Efficiency (${efficiencyLabel}): %{y:.2f}<extra></extra>`,
       }});
 
+    const maxEfficiency = Math.max(...chillerOutput.flatMap(c => c.ariEfficiencyProfile.slice(1)));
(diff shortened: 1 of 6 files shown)


src/app/process-cooling-assessment/services/process-cooling-charts.service.ts
```diff
@@ -28,6 +28,11 @@ export class ProcessCoolingChartsService {
         hovertemplate: `${chiller.name}<br>Load: %{x}<br>Efficiency (${efficiencyLabel}): %{y:.2f}<extra></extra>`,
       }});
 
+    const maxEfficiency = Math.max(...chillerOutput.flatMap(c => c.ariEfficiencyProfile.slice(1)));
```
nbintertech, line 31: Tickstep is now dynamic, not just the max tick. Fine with me, but PC experts might want a constant scale/step. You could toss a screen by Kristina to get an early check.

Let's get some commenting on tickStep to explain what that's doing (i.e. what does 6 represent? tick distribution? explain the ternary).

nbintertech (approved):

nbintertech (approved):

Issue #8403, opened 2026-04-29
Water Color Palettes Part Deux: Assessment feature


nbintertech:
This is the assessment side of #8187 and should be started after that issue is merged to develop. I'm putting in some guide information below but you should update this issue to your own requirements after you investigate

**Describe the feature**
Ensure user selected color palettes from the diagram are applied and persisted for any existing UI areas where they are used in the water assessment. 

Consider implicit requirements

If team thinks we need a color selector on the assessment side, this will warrant a new issue because we'll need to share/change the integration to some degree.

**Describe MEASUR location**
Water assessment.

**Describe MEASUR UI**
Sweep through the assessment and update with areas/UI

**Who should verify**

---
### Optional
**Feature requested by**
Kiran through training feedback

Issue #8187, opened 2026-02-12
Allow user to select global color palette


nbintertech:
The color management in diagram needs an upgrade before this can be done, TBD issue

**Describe the feature**
Users should be able to select a default color palette that corresponds to water components (I.E. water using system, waste water treatment, intakes)

**Describe MEASUR location**
Diagram left sidebar, below component drag and drop

**Describe MEASUR UI**
We've discussed having a dropdown with 5-10 options, each option showing a condensed version of the color palette. Should get consensus on palette options before beginning

I brought up the idea that we could just as well have them customize the color for each (and every instance) of the component, but the above solution was preffered.

**Who should verify**
Kiran/Kirstina

---
### Optional
**Feature requested by**
Kiran, Training feedback

nbintertech:
Discussed on 3/24:

<img width="2506" height="1740" alt="Image" src="https://github.com/user-attachments/assets/470f4277-56e8-4c74-8897-2e73f23093dc" />

Add in 5-10 palettes. 

Do some research to make them color blind or at minimum contrast ratio accessible. One example below:
https://davidmathlogic.com/colorblind/#%23D81B60-%231E88E5-%23FFC107-%23004D40

WoodrowKromar:
I attempted to use react-colorful but the color-picker functionality is not visible. It may be because I am using Edge, but colorful should be good with all browsers and has no dependencies. I will try again today after discussions.
Npm package: https://www.npmjs.com/package/react-colorful#browser-support

koay9f:
love this!

PR #8450, opened 2026-05-15
Issue 8403: True Cost Editable Table Pills pick up color palette changes

WoodrowKromar:
#8403
- After selecting a Color palette from the drop-down menu in Water Diagram -> options, those colors will reflect in the pill icons in the True Cost Editable Table (report -> True Cost Of Systems -> Adjust Cost Attribution.
- Removed old pill styles.

If at any point Known Loss is added, color palettes and mapping must be updated both here and in the diagram. See #8187 for locations to adjust.


WoodrowKromar:
image
-colors persist and maintain sync.

src/app/water/water-report/true-cost-editable-table/true-cost-editable-table.component.ts
```diff
@@ -213,6 +224,29 @@ export class TrueCostEditableTableComponent {
...
   }
 
+  getPillStyles(type: WaterProcessComponentType): { background: string; color: string } {
+    const idx = this.PALETTE_COMPONENT_ORDER.indexOf(type);
+    const bg = idx !== -1 && this.paletteColors[idx] ? this.paletteColors[idx] : undefined;
+    if (!bg) return { background: '#e0e0e0', color: '#000000' };
+    return { background: bg, color: this.getContrastTextColor(bg) };
+  }
+
+  private getContrastTextColor(bgColor: string): string {
```
nbintertech: Duplicate of method being used in the diagram already. Let's pull it out into process-flow-lib so it can be used as a helper in the diagram and the assessment

nbintertech (changes requested):
We should also add the default palette to userDiagramOptions in the json data of the water example in 'ExamplesData.json'

WoodrowKromar:
getContrastTextColor reduced to a single definition, and the userDiagramOptions in ExamplesData.json was given the default color palette.

nbintertech (approved):

Issue #8522, opened 2026-06-01
Steam Leak Survey Form


WoodrowKromar:
**Describe the feature**
For only creating the Survey Form Component Visual form of Steam Leak Calc. 
- html, .ts and form management with validation.
- use the Existing CA Leak Survey Component and structure for layout and form functionality.

Includes the 4 Forms.
- Form: basic form, the rest are sub-forms.
- Estimate-method form
- orifice-methid
- plume-method

DOES NOT INCLUDE: Frontend calculations.

PR #8539, opened 2026-06-03
Issue 8522: Steam Leak Forms, Functionality.

WoodrowKromar:
#8522 
- Steam System and Leak Data Forms.
- Major Service work for behavior and data manipulation.
- General behavior and Styling.
- Save operations, display of base forms, not results or help.


src/app/tools-suite-api/calculator-suite-api.service.ts
```diff
@@ -425,6 +427,65 @@ export class CalculatorSuiteApiService {
...
+      boilerEfficiency: facility.boilerEfficiency,
+      systemEfficiency: facility.systemEfficiency,
+      utilityType: utilityTypeMap[facility.utilityType] ?? 'electric',
+      fuelCost: facility.fuelCost,
+      fuelEnergyFactor: facility.fuelEnergyFactor,
+      steamCost: facility.steamCost,
+    };
+
+    switch (leak.measurementMethod) {
+      case 1:
```
nbintertech, line 460: Use the enum for these cases (like you're doing elsewhere). The less we have to remember how an int maps to a domain rule, the better,

src/app/calculator/steam/steam-leak/cost-of-steam-form/cost-of-steam-form.component.html
```diff
@@ -0,0 +1,190 @@
...
+      <div class="input-group">
+        <input name="electricityCost" class="form-control" type="number" step="any"
+          formControlName="electricityCost" id="steam-electricityCost" onfocus="this.select();"
+          (input)="save()" (focus)="changeField('electricityCost')">
+        <span class="input-group-addon units">$/kWh</span>
+      </div>
+    </div>
+  }
+
+  @if (facilityForm.controls.utilityType.value === 2) {
```
nbintertech, line 100: Similar as other comment, we can do an enum for the utility types also

nbintertech (approved):

Issue #8547, opened 2026-06-04
Steam Leak Estimate, Orifice, and Plume sub-forms.


WoodrowKromar:
Each form is displayed on the choice made for the measurement method. Each form is very similar with different fields.
Since they are so small, they will be one PR

PR #8558, opened 2026-06-09
Issue 8547: Round 2.Adding Missing fields, Adding Validation, Removing duplicates

WoodrowKromar:
#8547
- Adjusted Estimate, Orifice, and Plume forms with additional fields from the Calculator Template. Added adjustments to form logic in services.
- Centralized pressureReductionMethods and pass it as an input.
- Added Validation logic for fields and forms. Validation taken from the Calculator Template Document.
- I adjusted the Min value of Feedwater Temperature to 32 deg as its not Feedice Temperature. The document has a 0 deg constraint, but seems like a mistake.


nbintertech (approved):
By and large looks good. A few notes on the things (I believe) you're still in progress on:
- Keep validation message language for mins/maxes as closely as you can with what the app is doing in assessments
- If the min/max values came from some named domain constraint in the document, assign a const with the domain name, or include that in a comment with the static value passed to the validator. Example `const boilingPointOfWater = 212`

Issue #8524, opened 2026-06-01
Steam Calc Copy Table


WoodrowKromar:
**Describe the feature**
- Create Copy Table using CA Leak Survey Copy Table as a template.

**Verified By**
@nbintertech

PR #8559, opened 2026-06-09
Issue 8524: Copy Table Added

WoodrowKromar:
#8524 
- Added Steam Leak Calc Copy Table
- Tested copy results, layout and results are correct.


nbintertech (changes requested):
Doesn't appear to copy values after generating example, pasting snippet into an Excel blank sheet. Also has table columns that don't exist in the leak table

Image

nbintertech (changes requested):
Their are still additional columns included in the copy data. If there are requirements for that, continue as you've done, otherwise just use what's in the table html so the columns are a 1:1 representation

nbintertech (approved):

PR #8793, opened 2026-08-12
Steam Leak Calculator

WoodrowKromar:
# Pull Request Template
Complete Steam leak Calculator

## Checklist
- Variable, function, and class names are descriptive
- Code leverages existing architecture, helper methods, and shared modules when applicable
- Development artifacts such as console logs, test values, and comments have been removed
- All Copilot, Claude Code, or other LLM implementations have been reviewed as if contributed by another developer
- UI changes have been manually tested for usability and appearance
- This PR references the related issue number (if applicable), ex. issue -> `#0000`

## Description
- **For development team**: Optional. Update the issue with a comment, if necessary.
- **For open source contributors**: The PR description clearly explains the purpose and scope of the changes

---


WoodrowKromar:
The list of recent requested modifications:

- (completed) "Fuel Energy Factor": Remove the field.

- (completed) "Steam Temperature" is before "Steam Pressure" in the system data inputs but "Header Pressure" is before "Header Temperature" in the leak data.  Change the order for the header inputs to match the order for the system inputs?

- (completed) If the user specifies their system utility type as anything other than electricity, we will need to add a "cost of electricity" field to the leak data inputs that would get tied to the costOfElectricity variable in the suite. 

- (Completed) UI inputs need to be checked for the rest of the leak methods (orifice and plume). The plume method seems to still be pulling "system steam temperature" and not the "header steam temperature".

- (completed) When the use selects "None" as the pressure reduction method, set the header temperature and pressure equal to the system temperature and pressure.

- (completed) Default value for costOfElectricity is the same as settings.electricityCost.

nbintertech (approved):

